Issue #226, opened 2014-06-19
SecurityContext and Authentication behaviour in OAuthProviderProcessingFilter is incompatible with DelegatingSecurityContext* classes


autayeu:
DelegatingSecurityContextCallable and DelegatingSecurityContextRunnable are useful to provide SecurityContext for a thread in a convenient manner. However, current behaviour of the OAuthProviderProcessingFilter prevents them being used as intended.

Namely, the Delegating\* classes keep the context, while the provider replaces authentication in that specific context. That leads to access denied exceptions due to Authentication being replaced for the wrong one.

My understanding that for the purposes of keeping-restoring-clearing the security context the whole of SecurityContext should be used, not just authentication part of it.

Example:
1) Working thread schedules a separate thread to get some auxiliary work done.
2) Auxiliary thread inherits the security context by using DelegatingSecurityContext\* classes
3) Working thread finishes the processing and in OAuthProviderProcessingFilter.resetPreviousAuthentication the authentication is restored to the previous one (often anonymous) and that changes both working and auxiliary thread.
4) Auxiliary thread has access control problems.

OAuthProviderProcessingFilter restores authentication directly, instead of restoring the context, and that leads to authentication being changed in the auxiliary thread as well, thus defeating the purpose of DelegatingSecurityContext*.

I would change the filter to save-restore-clear security context instead of just authentication. Would such a patch be considered for acceptance or current behaviour has other reasons behind it and the DelegatingSecurityContext\* classes should have "siblings" doing DelegatingAuthentication\* stuff in addition to context?

PR #246, opened 2014-08-22
OAuth 1 provider refactoring from filters to end points (a-la OAuth2)

autayeu:
Github-226: https://github.com/spring-projects/spring-security-oauth/issues/226
# OAuth refactoring notes
## Migration for users with namespace configuration.
- Check security of OAuth end point URLs (most likely in applicationContext.xml, <http\intercept-url> elements). See sparklr/applicationContext.xml example.
  - request and access token endpoints need to be accessible to the consumer (they handle requests from the consumer (server), most likely authenticated only by OAuth signature)
  - access confirmation and authorization end points need to be restricted to authorized users, most likely ROLE_USER or similar. They handle requests from the user (browser) and authenticated by standard Spring Security mechanism.
## Migration for users with manual bean configuration.
- Replace old filters with one OAuthProviderProcessingFilter (usually after ExceptionTranslationFilter) - it should process protected resources and at least request token and access token end point URLs.
- Check that FrameworkEndpointHandlerMapping registers end points URLs and they are properly and not overly protected by the interceptors (see above).
- Use FrameworkEndpointHandlerMapping.setMappings to customize end point URLs. See OAuthConstants.java for default URLs. Example:

``` xml
    <bean id="frameworkEndpointHandlerMapping" class="org.springframework.security.oauth.provider.endpoint.FrameworkEndpointHandlerMapping">
        <property name="mappings">
            <util:map>
                <entry key="/oauth_request_token" value="/oauth/request_token"/>
                <entry key="/oauth_access_token" value="/oauth/access_token"/>
                <entry key="/oauth_authenticate_token" value="/oauth/authorize_token"/>
            </util:map>
        </property>
    </bean>
```
- Wire beans to OAuthProviderProcessingFilter (e.g. tokenServices, consumerDetailsService, authHandler, frameworkEndpointHandlerMapping). Wire beans to end points: tokenServices for all; rememberMeServices, applicationEventPublisher, verifierServices, authenticationSuccessHandler, authenticationFailureHandler for AuthorizationEndpoint; if needed, configure require10a where present.
## Localization messages key changes

Messages have changed. The changed, removed and added keys follow (old = new):

```
AccessTokenProcessingFilter.missingToken = AccessTokenEndpoint.missingToken
AccessTokenProcessingFilter.missingVerifier = AccessTokenEndpoint.missingVerifier
AccessTokenProcessingFilter.missingVerifier =  AccessTokenEndpoint.invalidVerifier
AccessTokenProcessingFilter.timestampNotNew = removed

OAuthProcessingFilter.missingConsumerKey = OAuthProviderProcessingFilter.missingConsumerKey
OAuthProcessingFilter.missingCredentials = OAuthProviderProcessingFilter.missingCredentials
OAuthProcessingFilter.incorrectRealm = OAuthProviderProcessingFilter.incorrectRealm
OAuthProcessingFilter.missingSignatureMethod = OAuthProviderProcessingFilter.missingSignatureMethod
OAuthProcessingFilter.missingSignature = OAuthProviderProcessingFilter.missingSignature
OAuthProcessingFilter.missingTimestamp = OAuthProviderProcessingFilter.missingTimestamp
OAuthProcessingFilter.missingNonce = OAuthProviderProcessingFilter.missingNonce
OAuthProcessingFilter.invalidTimestamp = OAuthProviderProcessingFilter.invalidTimestamp
OAuthProcessingFilter.timestampNotNew = removed

ProtectedResourceProcessingFilter.missingToken = OAuthProviderProcessingFilter.missingToken

AccessTokenProcessingFilter.missingCallback = RequestTokenEndpoint.missingCallback
new = RequestTokenEndpoint.missingCredentials
```
### Architecture

Upgraded provider architecture: one filter (OAuthProviderProcessingFilter) and several end points (AbstractEndpoint subclasses, standard Spring MVC request mappings). OAuthProviderProcessingFilter processes signatures and loads authentication for framework end points and for protected resource requests.
### Code
- OAuthParameter.java - merged consumer and provider parameters, because they were used interchangeably.
- OAuthParameters.java - container for OAuth requests. POJO is faster than map.
- StringSplitUtils.java - refactored to use List instead of array to reduce memory allocations.
- OAuthProviderBeanDefinitionParser.java - upgraded to configure the refactored framework.
- CoreOAuthConsumerSupport.java - refactoring collateral.
- OAuthConsumerContextFilter.java - refactoring collateral.
- AbstractEndpoint.java - base class for framework end points.
- AccessTokenEndpoint.java - access token end point.
- AuthorizationEndpoint.java - token authorization end point.
- FrameworkEndpoint.java - annotation to mark framework end points. Includes a parameter to differ the ones not requiring OAuth signature.
- FrameworkEndpointHandlerMapping.java - handler mapping which sets up the end points.
- RequestTokenEndpoint.java - request token end point.
- UserAuthorizationSuccessfulAuthenticationHandler.java - moved.
- CoreOAuthProviderSupport.java - refactoring collateral.
- OAuthProviderProcessingFilter.java - core filter. Now a OncePerRequestFilter, processes OAuth signatures and loads consumer and user authentication when possible. Changes:
  - removed allowedMethods. Now controlled by endpoints.
  - removed filterProcessesUrl. Now protects both end points and resources.
  - changed SecurityContext processing. Now a new context is established before OAuth processing and the whole context is saved-restored, not just Authentication.
- ConsumerAuthentication.java - refactoring collateral.
- OAuthProviderSupport.java - refactoring collateral.

There is a new dependency on spring-webmvc.
### Tests
- AuthorizationServerBeanDefinitionParserTests.java - refactoring collateral damage.
- CoreOAuthConsumerSupportTests.java - refactoring collateral damage.
- OAuthConsumerContextFilterTests.java - refactoring collateral damage.
- AccessTokenEndpointTests.java - increased code coverage.
- AuthorizationEndpointTests.java - increased code coverage.
- RequestTokenEndpointTests.java - increased code coverage.
- UserAuthorizationSuccessfulAuthenticationHandlerTests.java - refactoring collateral damage.
- OAuthProviderProcessingFilterTests.java - increased code coverage.
- CoreOAuthProviderSupportTests.java - increased code coverage.

Debug output is enabled in tests to increase code coverage and test debug clauses (construction of messages, etc).


alexzm1:
Hi guys this looks great, is there any chance we can get these changes in one of our close releases? Probably I'm sonehow lost in the release plan of Spring-security-oauth but, when is the next release scheduled to happen?

dsyer:
This ia a pretty major change, so it is going to have to be a 2.1 or a 3.0 release (no plans currently, but if you try this patch and it definitely works, that gives us a good reason to plan it).

autayeu:
By the way, Alex, Dave, I'm around if need be to ask questions on this ;)

autayeu:
I've updated the branch to resolve merge conflicts introduced by ```<br>``` removal in JavaDocs.

alexzm1:
I'm working on an oauth app that uses spring-security-oauth, I can test those changes in my app and report any issue that I can find with this new patch, are there any specific test cases that we have to follow besides the common oauth flow that we have to test?

spring-security-oauth/src/main/java/org/springframework/security/oauth/provider/filter/CoreOAuthProviderSupport.java
```diff
@@ -1,268 +1,310 @@
...
+	public String getSignatureBaseString(HttpServletRequest request) {
+		SortedMap<String, SortedSet<String>> significantParameters = loadSignificantParametersForSignatureBaseString(request);
 
-  // Inherited.
-  public String getSignatureBaseString(HttpServletRequest request) {
-    SortedMap<String, SortedSet<String>> significantParameters = loadSignificantParametersForSignatureBaseString(request);
+		// now concatenate them into a single query string according to the spec.
+		StringBuilder queryString = new StringBuilder();
+		Iterator<Map.Entry<String, SortedSet<String>>> paramIt = significantParameters.entrySet().iterator();
+		while (paramIt.hasNext()) {
```
alexzm1: Is there any reason we're not iterating this with a for each statement?

spring-security-oauth/src/main/java/org/springframework/security/oauth/provider/filter/CoreOAuthProviderSupport.java
```diff
@@ -1,268 +1,310 @@
...
+	 * The parameters will be encoded, per the spec section 9.1.
+	 *
+	 * @param request The request.
+	 * @return The significant parameters.
+	 */
+	protected SortedMap<String, SortedSet<String>> loadSignificantParametersForSignatureBaseString(HttpServletRequest request) {
+		// first collect the relevant parameters...
+		SortedMap<String, SortedSet<String>> significantParameters = new TreeMap<String, SortedSet<String>>();
+		// first pull from the request...
+		Enumeration parameterNames = request.getParameterNames();
```
alexzm1: I was thinking that we can change from 195 to 217 for this

```java
for(Entry<String, String[]> parameter : request.getParameterMap().entrySet()){
	String parameterName = oauthEncode(parameter.getKey());
	for (String parameterValue : parameter.getValue()) {
		if (parameterValue == null) {
			parameterValue = "";
		}
		parameterValue = oauthEncode(parameterValue);
				
		if(!significantParameters.containsKey(parameterName)){
			significantParameters.put(parameterName, new TreeSet<String>());
		}
		significantParameters.get(parameterName).add(parameterValue);
	}
}
```
That way we can avoid using the `Enumeration` to iterate through the parameters

spring-security-oauth/src/main/java/org/springframework/security/oauth/provider/filter/CoreOAuthProviderSupport.java
autayeu: Perhaps because it is easier to check whether we need to add '&':
```java
				if (paramIt.hasNext() || valueIt.hasNext()) {
					queryString.append('&');
				}
```
Alternative with "for each" would be to add '&' anyway, then check and delete after the end of the loop.

spring-security-oauth/src/main/java/org/springframework/security/oauth/provider/filter/CoreOAuthProviderSupport.java
autayeu: Yes, it is possible. But there are two things that can be improved. The first is duplicated map lookup:
```java
            if(!significantParameters.containsKey(parameterName)){
                significantParameters.put(parameterName, new TreeSet<String>());
            }
            significantParameters.get(parameterName).add(parameterValue);
```
```containsKey``` - one lookup, ```get``` - second lookup. Current version has just one lookup. And the second is that the provided piece of code breaks google compatibility test.

Comment (alexzm1) [inline spring-security-oauth/src/main/java/org/springframework/security/oauth/provider/filter/CoreOAuthProviderSupport.java]: Yeah you're right that's two lookups it could improve readability but it is at expenses of performance, I'm curios about the second concern, What is the Google Compatibility Test?

File: spring-security-oauth/src/main/java/org/springframework/security/oauth/provider/filter/CoreOAuthProviderSupport.java [hunk 3]
```diff
@@ -1,268 +1,310 @@
...
  * @author Ryan Heaton
+ * @author <a rel="author" href="http://autayeu.com/">Aliaksandr Autayeu</a>
  */
 public class CoreOAuthProviderSupport implements OAuthProviderSupport {
 
-  private final Set<String> supportedOAuthParameters;
-  private String baseUrl = null;
+	private String baseUrl = null;
+
+	public CoreOAuthProviderSupport() {
```
alexzm1: Do we really need to define this constructor?

spring-security-oauth/src/main/java/org/springframework/security/oauth/common/OAuthParameter.java
```diff
@@ -17,11 +17,12 @@
 package org.springframework.security.oauth.common;
 
 /**
- * Enumeration for consumer parameters.
+ * OAuth parameters.
  *
  * @author Ryan Heaton
+ * @author <a rel="author" href="http://autayeu.com/">Aliaksandr Autayeu</a>
  */
-public enum OAuthConsumerParameter {
```
alexzm1: I know that this enum was here before your chances but, I'm not quite sure why they decided to make this an Enum if they are only making reference to this enum class to get their constants, I can't find any reference in the code where they are storing an ``OAuthConsumerParameter`` or a ``OAuthParameter`` type in any place, it seems that all the places in the project where they made reference to this enum is to retrieve one of it's enum instances and made a `Enum.toString()` call, why instead of having the enum we don't just made this a constants class, that way we can call the public constant and we can save the `.toString()` call in the code of those places we made reference to this.

alexzm1: ... It seems that there is a place where we actually need this to be an Enum, they were actually iterating through the Enum values in class `org.springframework.security.oauth.consumer.client.CoreOAuthConsumerSupport` line 365, so disregard my previous comment.

spring-security-oauth/src/main/java/org/springframework/security/oauth/common/OAuthParameters.java
```diff
@@ -0,0 +1,155 @@
...
+package org.springframework.security.oauth.common;
+
+import java.io.Serializable;
+
+/**
+ * Container for parsed OAuth parameters.
+ *
+ * @author <a rel="author" href="http://autayeu.com/">Aliaksandr Autayeu</a>
+ */
+public class OAuthParameters implements Serializable {
```
alexzm1: We're missing the `serialVersionUID`

autayeu: Thank you, I've added it.

spring-security-oauth/src/main/java/org/springframework/security/oauth/provider/filter/CoreOAuthProviderSupport.java
autayeu: I apologize for ambiguity, I mean ```GoogleCodeCompatibilityTests.java```. Replace the code with your proposal, run tests and you'll see it breaks. I didn't explore the details further.

alexzm1:
Hi guys, it's been 6 months since we worked on this Pull Request, is there something that I can help to have these changes in a release? I would really like to have those changes in a stable release.

Regards.

dsyer:
Thanks for the reminder. The only thing that stops work on 2.1 is really the amount of stuff still happening in 2.0. If you want to speed things up you could pick up issues and send pull requests for 2.0.8, but since most of the open issues are about OAuth2 you might not be interested in doing that.

jgrandja:
@autayeu Are you looking to still get this merged? If so, I'm quite reluctant on merging this since it touches on too many files. My preference would be to split this PR into more manageable updates. We have 2.5.0 scheduled for May 7 so we can look at merging for that release if you're able to split up this PR. Let me know your thoughts.

autayeu:
@jgrandja I'm sorry, I don't have bandwidth.

jgrandja:
@autayeu No worries. I'll close the PR then. Feel free to submit a new PR later on if you still require the changes.

Issue #330, opened 2014-12-08
Token endpoint should permit all HTTP OPTIONS requests to support CORS.


ccampo133:
This is the same issue documented in https://github.com/spring-projects/spring-security-oauth/issues/184. That ticket was closed and marked as fixed, but I believe this fix was removed in subsequent updates. The fix was supposedly committed in 8d1e947545adad7524f7a5b173a3067efad6c52f, yet removed in d3819e2e5414019ab9e5cb1207fc6a54f239dd55. Was this intentional? If so, why?

greyfairer:
FYI the solutions in http://docs.spring.io/spring/docs/current/spring-framework-reference/html/cors.html do not work, because the token endpoint is intercepted before it reaches the MVC handler.

The workaround I found using Spring Boot 1.3 and Spring Security OAuth2 2.0.8 would be to remove @EnableAuthorizationServer, and supply a overridden AuthorizationServerSecurityConfiguration as below:

```
//workaround for https://github.com/spring-projects/spring-security-oauth/issues/330
@Configuration
@Import(AuthorizationServerEndpointsConfiguration.class)
public class CorsEnabledAuthorizationServerSecurityConfiguration extends AuthorizationServerSecurityConfiguration {

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        super.configure(http);
        CorsConfigurationSource source = corsConfigurationSource();
        http.addFilterBefore(new CorsFilter(source), ChannelProcessingFilter.class);
    }

    private CorsConfigurationSource corsConfigurationSource() {
        UrlBasedCorsConfigurationSource source = new UrlBasedCorsConfigurationSource();
        CorsConfiguration config = new CorsConfiguration();
        config.addAllowedOrigin("*");
        config.addAllowedHeader("*");
        config.addAllowedMethod("POST");
        //more config
        source.registerCorsConfiguration("/**", config);
        return source;
    }
}
```

ccampo133:
Yea I noticed that Spring Boot 1.3 broke CORs support for Spring OAuth2. I
too have to implement a custom authorization configuration. Would be nice
if this ticket was reopened and reevaluated.

dsyer:
This ticket is still open, but has a pretty low priority. I don't think you need any custom security configuration (the CORS filter is not a Security filter, and I already remarked in an earlier comment). You can just add one with a path that suits you (e.g. matching only the token endpoint).

greyfairer:
[...]

Issue #327, opened 2014-12-02
Only allow HTTP POST to token endpoint by default


ccampo133:
Per the OAuth2 spec, requests to the token endpoint should only use the HTTP POST request method.

[Section 3.2](http://tools.ietf.org/html/rfc6749#section-3.2):
> ... 
The client MUST use the HTTP "POST" method when making access token requests.
...

Currently, Spring Security OAuth2 allows you to get request tokens using all the HTTP verbs (GET, PUT, PATCH, POST, and even DELETE) by default! This endpoint should be restricted to HTTP POST by default instead.

This should be fairly simple to implement; just change the `@RequestMapping(value = "/oauth/token")` annotation on `TokenEndpoint.getAccessToken` to `@RequestMapping(value = "/oauth/token", method = RequestMethod.POST)`.

The workaround for this wasn't trivial. I had to provide a separate `WebSecurityConfigurerAdapter` which explicitly denied the verbs other than POST, with `@Order(-1)` so it was favored over the default Spring configurations. Also, unless I'm missing something, Spring Security doesn't allow one to set multiple matchers for the same endpoint, but with different request methods (i.e. separate ant matchers for '/oauth/token' GET, PUT, DELETE, etc.), so this had to be done by instantiating new `AntPathRequestMatcher`s directly:

    @Configuration
    @EnableWebSecurity
    @Order(-1)
    protected static class CustomTokenSecurityConfig extends WebSecurityConfigurerAdapter {

        @Override
        protected void configure(final HttpSecurity http) throws Exception {
            // @formatter:off
            http
                .requestMatchers()
                    .requestMatchers(
                            new AntPathRequestMatcher("/oauth/token", HttpMethod.GET.toString()),
                            new AntPathRequestMatcher("/oauth/token", HttpMethod.PUT.toString()),
                            new AntPathRequestMatcher("/oauth/token", HttpMethod.PATCH.toString()),
                            new AntPathRequestMatcher("/oauth/token", HttpMethod.DELETE.toString()))
                .and()
                .csrf()
                    .disable()
                .authorizeRequests()
                    .anyRequest()
                        .denyAll();
            // @formatter:on
        }
    }

Alternatively, I can also extend `AuthorizationServerSecurityConfiguration` and override `configure(HttpSecurity http)` there, (and import it, rather than use `@EnableAuthorizationServer`) like so:

    @Configuration
[...]

Issue #184, opened 2014-04-15
CORS problem


raonirenosto:
I enabled CORS on my project, then I noticed that HTTP OPTIONS method is being authenticated. I'm using java configuration, and I'm not able to reconfigure endpoints like /oauth/token to allow the OPTIONS method. I changed the native code as shown:

```
    String tokenEndpointPath = oauth2EndpointHandlerMapping().getPath("/oauth/token");
    // @formatter:off
    // allow option method
    http.authorizeRequests().antMatchers(HttpMethod.OPTIONS, tokenEndpointPath).permitAll();
    http
    .authorizeRequests()
        .antMatchers(tokenEndpointPath).fullyAuthenticated()
        .and()
    .requestMatchers()
        .antMatchers(tokenEndpointPath);
    // @formatter:on
```

Now is working as expected. Is this a problem/bug, or I could redefine this endpoint without changing native code?

PR #331, opened 2014-12-10
Add permitAll() to OPTIONS requests on token endpoint to support CORS and restrict token endpoint to HTTP POST.

ccampo133:
Added permitAll() for OPTIONS requests back to the token endpoint to fix https://github.com/spring-projects/spring-security-oauth/issues/330 and https://github.com/spring-projects/spring-security-oauth/issues/184.

Also, restricted token endpoint to only allow HTTP POST per https://github.com/spring-projects/spring-security-oauth/issues/327.


spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/annotation/web/configuration/AuthorizationServerSecurityConfiguration.java
```diff
@@ -68,6 +69,9 @@ protected void configure(HttpSecurity http) throws Exception {
 		String checkTokenPath = handlerMapping.getServletPath("/oauth/check_token");
 		// @formatter:off
 		http
+			.authorizeRequests()
```
kopax: @ccampo133 what is the issue id please

PR #334, opened 2014-12-11
Restricted token endpoint to HTTP POST by default.

ccampo133:
Restricted the token endpoint to only allow HTTP POST requests by default. Proposed solution to https://github.com/spring-projects/spring-security-oauth/issues/327

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/annotation/web/configuration/AuthorizationServerEndpointsConfiguration.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/annotation/web/configuration/AuthorizationServerEndpointsConfiguration.java
@@ -15,6 +15,7 @@ package org.springframework.security.oauth2.config.annotation.web.configuration;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 
 import javax.annotation.PostConstruct;
 
@@ -28,6 +29,7 @@ import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProce
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
+import org.springframework.http.HttpMethod;
 import org.springframework.security.oauth2.config.annotation.web.configuration.AuthorizationServerEndpointsConfiguration.TokenKeyEndpointRegistrar;
 import org.springframework.security.oauth2.config.annotation.web.configurers.AuthorizationServerEndpointsConfigurer;
 import org.springframework.security.oauth2.provider.ClientDetailsService;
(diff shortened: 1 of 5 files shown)


dsyer:

ccampo133:
Yessir :)

vkhoroshko:
Could someone tell us how to override allowed methods if we're using <oauth:authorization-server in XML and not @EnableAuthorizationServer?

Thanks in advance

dsyer:
What is your use case for requiring anything but a POST? I'm mainly interested because I think we should only allow POST.

dsyer:
@vkhoroshko you can probably quite easily inject the `TokenEndpoint` into one of your beans and set the `allowedRequestMethods`.

vkhoroshko:
The issue is that our clients are already using **GET** and we just can't break them.

For now we created post processor that sets `allowedRequestMethods`.

But, anyway, I believe this pull request is a little bit incomplete in terms that if something is allowed to be configured using class based configuration - the same should be allowed for XML configuration also.

ccampo133:
Considering the OAuth 2 spec says that clients are **only** allowed to use POST, I think it is reasonable for an OAuth 2 library to match the spec. That being said, it is true that you can do this in Java config and not XML, but that can be said of quite a few things in Spring. Since there is a workaround, if this is even considered an issue, it should be a low priority.

michael-simons:
In case anyone comes looking for this:

```
import java.util.HashSet;
import java.util.Set;

import org.springframework.beans.BeansException;
import org.springframework.beans.factory.config.BeanPostProcessor;
import org.springframework.http.HttpMethod;
import org.springframework.security.oauth2.provider.endpoint.TokenEndpoint;

public class TokenEndpointPostProcessor implements BeanPostProcessor {

	@Override
	public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
		if (bean instanceof TokenEndpoint) {
			final TokenEndpoint tokenEndpoint = (TokenEndpoint) bean;
			final Set<HttpMethod> allowedRequestMethods = new HashSet<>();
			allowedRequestMethods.add(HttpMethod.POST);
			allowedRequestMethods.add(HttpMethod.GET);
			tokenEndpoint.setAllowedRequestMethods(allowedRequestMethods);
		}
		return bean;
	}

	@Override
	public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
		return bean;
	}
}
```

@dsyer If i find the time to add this to the xml parser, any chance of merging it?

pivotal-issuemaster:
@ccampo133 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=334)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=334) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ccampo133 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=334)!

Issue #351, opened 2015-01-07
AuthorizationServerEndpointsConfigurer has the wrong default for reuseRefreshToken


dsyer:
It looks like the change in commit 793ebe0a made a false assumption about the default value in `DefaultTokenServices`.

PR #352, opened 2015-01-08
#351: added setter for reuseRefreshToken in AuthorizationServerEndpoints...

jonasgeiregat:
Added setter to AuthorizationServerEndpointsConfigurer which makes the reuse of the refresh token flexible. Choosen this above renaming the method to doNotReuseRefreshToken or setting the default value of reuseRefreshToken to false which would conflict the default value of DefaultTokenServices.


Trundle:
Commit bdb3390b19c54bdca3da6383ec9ad191ef4734cd makes this pull request obsolete.

dsyer:
Fixed already. Thanks.

Issue #362, opened 2015-01-16
OAuth2ExpressionParser class should be public to be instantiable


paskos:
`org.springframework.security.oauth2.provider.expression.OAuth2ExpressionParser` has a public constructor but is currently package protected.

I'd like to use it in my project but without using the associated OAuth2 `AbstractSecurityExpressionHandler` subclasses.

right now I have a hack where my code is in the same `org.springframework.security.oauth2.provider.expression` package but it would be nice if I simply could instantiate it from outside that package.

paskos:
Pull request can be found [here!](https://github.com/spring-projects/spring-security-oauth/pull/365)

dsyer:
What's your use case? I don't think it's always a good idea to make internal implementation details public.

paskos:
I work on a large project where security access rules are dynamic (not in config classes or files), complex (not role based) and shared among many "protocols" (http only, oauth2 over http, jms, websocket).

We wrote a generic `org.springframework.security.access.expression.SecurityExpressionHandler` usable in all those protocol with very little customization (overriding of some targeted methods) in "protocol" sub-classes (PureHttp, OAuth2, SimpMessage....).

One of those customization is the ability to provide a "native" (not implemented by us) `org.springframework.security.oauth2.provider.expression.ExpressionParser`.
 `OAuth2ExpressionParser` being one of them because we like ` "#oauth2.throwOnError(" + expressionString + ")";`).

Unfortunately `OAuth2ExpressionParser` is package protected so our `OAuth2SecurityExpressionHandler`has to be declared in `org.springframework.security.oauth2.provider.expression` to be able to see it.

dsyer:
That's useful to know. More code would help though. Is there any reason you can't use the public API (`OAuth2WebSecurityExpressionHandler`)?

paskos:
`OAuth2WebSecurityExpressionHandler extends DefaultWebSecurityExpressionHandler` which assumes `FilterInvocation` is used.

That forces `OAuth2MethodSecurityExpressionHandler` to be used when `MethodInvocation` is the "security context".

`OAuth2MethodSecurityExpressionHandler` and `OAuth2WebSecurityExpressionHandler` both assume you will need `OAuth2SecurityExpressionMethods` which we don't in our case because we have other security methods.

[...]

PR #365, opened 2015-01-19
ISSUE #362

paskos:
OAuth2ExpressionParser class should be public to be instantiable

I work on a project with complex dynamic security access rules.
We implemented our own SecurityExpressionHandler where we want to
instantiate OAuth2ExpressionParser but can't because it's package
protected.


dsyer:
Fixed per discussion in #362. Thanks.

PR #402, opened 2015-02-13
Support “out of band” callback

dai0304:
Redirect urn "urn:ietf:wg:oauth:2.0:oob" is for Out-Of-Band authentication (where the app cannot be reached by the authenticating BROWSER)
With this value, authorization server redirect the user to a special authorization result page that supports two use cases:

First, if your application can detect changes to the title bar of the browser window, the result of authorization request can be retrieved automatically. An authorization server sets the document title and that is typically available in the title bar of the browser window.

If the authorization request was approved, the title will look like this:

```
Success code=xxxxxx&state=yyyyyyyy
```

If the authorization request was denied or failed, the title will look like this:

```
Denied error=access_denied&error_description=User%20denied%20access&state=yyyyyyyy
```

Once your application has detected the change to the document title and successfully retrieved the code, the browser window can be closed automatically if possible.

When using this approach, it is important to limit the state parameter to a reasonable length to reduce the chance of truncation. We suggest no more than 10 characters.

Second, as a fallback solution, the authorization code will also be printed to the screen so that the user can manually copy and paste it into the client application.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/AuthorizationEndpoint.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/AuthorizationEndpoint.java
@@ -82,11 +82,14 @@ import org.springframework.web.util.UriComponentsBuilder;
  * 
  * @author Dave Syer
  * @author Vladimir Kryachko
+ * @author Miyamoto Daisuke
  * 
  */
 @FrameworkEndpoint
 @SessionAttributes("authorizationRequest")
 public class AuthorizationEndpoint extends AbstractEndpoint {
+	
+	private static final String OOB_REDIRECT_URI = "urn:ietf:wg:oauth:2.0:oob";
 
 	private AuthorizationCodeServices authorizationCodeServices = new InMemoryAuthorizationCodeServices();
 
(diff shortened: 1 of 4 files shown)


tomvandenberge:
This feature would be very welcome.

tomvandenberge:
There's one thing missing to get the "out of band" endpoint working. It is not registered as a bean, and therefore not picked up. To do so, add the following to `AuthorizationServerEndpointsConfiguration`:

```
@Bean
public WhitelabelOutOfBandRedirectEndpoint whitelabelOutOfBandRedirectEndpoint() {
	return new WhitelabelOutOfBandRedirectEndpoint();
}
```

pivotal-issuemaster:
@dai0304 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=402) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@dai0304 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=402)!

jgrandja:
@dai0304 I'm not familiar with `urn:ietf:wg:oauth:2.0:oob`. Can you point me to the specification that details this?

jgrandja:
@dai0304 I'm closing this PR due to lack of feedback.

Issue #413, opened 2015-02-24
TokenEndPoint problem bind HttpMethod parameter


oassuncao:
In TokenEndPoint class on version 2.0.6 is added the functionality to support others httpMethods, and receive a parameter with HttpMethod. I'm using Spring 3 and in this version the Spring can't bind this parameter and the value is NULL and always I receive the exception HttpRequestMethodNotSupportedException, See the code below:

```
@RequestMapping(value = "/oauth/token")
	public ResponseEntity<OAuth2AccessToken> getAccessToken(Principal principal, @RequestParam
	Map<String, String> parameters, HttpMethod requestMethod) throws HttpRequestMethodNotSupportedException {

		if (!allowedRequestMethods.contains(requestMethod)) {
			throw new HttpRequestMethodNotSupportedException(requestMethod.toString());
		}

               ....
}
```

dsyer:
Spring 3 doesn't bind `HttpMethod` in a `@RequestMapping`?

oassuncao:
Yes, you can see this information in the documentation, Only in Spring 4 you can bind HttpMethod
Spring 3
http://docs.spring.io/spring/docs/3.2.13.RELEASE/spring-framework-reference/html/mvc.html#mvc-ann-arguments

Spring 4
http://docs.spring.io/spring/docs/current/spring-framework-reference/html/mvc.html#mvc-ann-arguments

dsyer:
I guess Spring 3 is not a dependency of this project, so it' snot really a surprise it doesn't work. The easiest way to get it to work is to simply not support anything other than POST anyway (which ironically is what @ccampo133 wanted in #327). 

I don't think we really want to commit to always supporting old versions of Spring, but this on the face of it a small thing to change (even though it might have a big effect on people who are misusing the token endpoint). Anyone else got opinions?

oassuncao:
To resolve my problem I do this workaround bellow. In the version 2.0.5 the Spring OAuth works fine with Spring 3.

```
        @Override
        public void configure(AuthorizationServerEndpointsConfigurer endpoints) throws Exception {
            ...
            endpoints.allowedTokenEndpointRequestMethods().getAllowedTokenEndpointRequestMethods().add(null);
        }
```

ccampo133:
[...]

PR #418, opened 2015-02-26
Switched from HttpMethod to HttpServletRequest in TokenEndpoint#getAccessToken

ccampo133:
Switched from HttpMethod to HttpServletRequest in `TokenEndpoint#getAccessToken`, as HttpMethod is not supported as a controller method argument until Spring 4, and we want to at least be able to support Spring 3. Addresses issue #413.


ccampo133:
Looks like this is taken care of in 39f605590856ccdd59f21a77f097331f098ea5c4. Closing this PR.

Issue #408, opened 2015-02-17
HTTP requests to the token endpoint other than POST return HTTP 500 instead of HTTP 405


ccampo133:
As part of Spring Security OAuth2 2.0.6, the token endpoint prohibits all request methods other than POST by default. Any requests other than POST to the token endpoint *should* return an HTTP status 405 "Method Not Allowed". Instead, HTTP 500 "Server Error" is returned.

This "bug" was actually introduced by my pull request (PR #334) to initially limit the token endpoint to POST by default. I'll see if I can create another PR to fix this behavior.

dsyer:
Thanks. That would be great.

PR #419, opened 2015-02-26
Added HTTP 405 response to TokenEndpoint

ccampo133:
Added HTTP 405 response from the token endpoint for unsupported HTTP request methods. Addresses issue #408.


dsyer:
I think we have this covered now.

PR #464, opened 2015-04-28
OAuth Token Introspect

beckje01:
Add a introspect endpoint for oauth.

Targeted at https://tools.ietf.org/html/draft-ietf-oauth-introspection-08

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/annotation/web/configuration/AuthorizationServerEndpointsConfiguration.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/annotation/web/configuration/AuthorizationServerEndpointsConfiguration.java
@@ -38,13 +38,7 @@ import org.springframework.security.oauth2.provider.OAuth2RequestValidator;
 import org.springframework.security.oauth2.provider.TokenGranter;
 import org.springframework.security.oauth2.provider.approval.UserApprovalHandler;
 import org.springframework.security.oauth2.provider.code.AuthorizationCodeServices;
-import org.springframework.security.oauth2.provider.endpoint.AuthorizationEndpoint;
-import org.springframework.security.oauth2.provider.endpoint.CheckTokenEndpoint;
-import org.springframework.security.oauth2.provider.endpoint.FrameworkEndpointHandlerMapping;
-import org.springframework.security.oauth2.provider.endpoint.TokenEndpoint;
-import org.springframework.security.oauth2.provider.endpoint.TokenKeyEndpoint;
-import org.springframework.security.oauth2.provider.endpoint.WhitelabelApprovalEndpoint;
-import org.springframework.security.oauth2.provider.endpoint.WhitelabelErrorEndpoint;
+import org.springframework.security.oauth2.provider.endpoint.*;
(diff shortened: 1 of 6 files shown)


beckje01:
I wasn't sure if I should also add it to the example apps, just let me know if you would like that.

dsyer:
Thanks. Did you fill in the contributor's agreement (link in README)?

beckje01:
I have in the past for grails. I may not be listed on the correct project, if there is anything you need from me let me know.

dsyer:
That's good enough probably. Now maybe you can explain why this new endpoint isn't just an option on the existing `CheckTokenEndpoint`? They look almost identical, and the intent is certainly the same.

beckje01:
The main difference is the flow on failure where the token introspection spec has it return a payload of the failure but the status code is a 200. 

Where the check token endpoint will return a 400.

I would be happy to make this an option inside checktoken endpoint but by changing that flow you couldn't easily have both running. I also wasn't sure how these would both evolve in the long run if the intent is to move the CheckTokenEndpoint and RemoteTokenServices towards the token introspection draft then it makes a lot of sense to me to keep them together and have it be an option.

In our project we are using the check token endpoint with spring security oauth built in RemoteTokenServices along side the Introspection endpoint with a few vertx applications.  

Just let me know your preference and I'll make the needed changes. If it becomes an option would it also make sense to update the RemoteTokenServices or to fork it?

dsyer:
The `CheckToken` endpoint predates the introspection spec, but its intent is identical, so I think it should evolve to match the spec, rather than being a parallel implementation. To start with (2.0) we can make the default the old style, and then switch at some point. I think the error return wouldn't stop many existing clients using the endpoint in the proposed new form anyway.

beckje01:
Sure I can redo this PR as an option for CheckToken, I'll also update RemoteTokenServices to have an option as well. Does that sound like a good path forward?

dsyer:
Yes that sounds great. Thanks.

beckje01:
I was closing so I could do the pr as an option like we discussed I finally am getting back to this is there something you want from the state it's in today@

dsyer:
OK. Sorry, I hadn't seen the new PR. Assumed you closed it by mistake. If it's not a breaking change we can squeeze it into 2.0.10 which ought to be out soon now we have some new resources.

beckje01:
I'll have a new PR late tonight I'll close this one at that point.

pivotal-issuemaster:
@beckje01 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=464) to manually synchronize the status of this Pull Request.

public:
@beckje01 I guess this can be closed now? :)

ractive:
@beckje01: I'm curious what happened with that. Did you do another PR as mentioned in the comments?

beckje01:
@ractive I never got it submitted if there is interest I'm happy to get it back to a pr state

ractive:
If it's just laying around, waiting to be PRed I'd say go for it. 😄 
I just stumbled across this because we're tweaking the infos that get returned by the CheckTokenEndpoint and I was wondering if it would make sense to add the "proper" token introspection already. I've seen API gateways that do support that.

PR #467, opened 2015-04-30
Adding support for properly handling the HttpMethodNotSupported error

kelsin:
We ran into a weird issue with the latest version of oauth where GET requests to the token endpoint resulted in a plain text (not xml or json) error to the user of "Internal Server Error" instead of a pretty message about GET's not being allowed.

This pull request handles `HttpRequestMethodNotSupportedException` the same as your other exceptions. The current code throws which is not allowed in a `@ExceptionHandler` handler. This change handles it like the other exceptions.

Then in order to not have the exception reveal itself as a 500 Server error I built in support for catching `HttpRequestMethodNotSupportedException` and turning them into a new `BadRequest` `OAuth2Exception`

The unit tests for `TokenEndpoint` aren't changed and there are no tests for `DefaultWebResponseExceptionTranslator` at the moment. I couldn't find the right place to add any type of integration tests (this is my first look into spring-security-oauth2 code) so if there is a proper place to add something please let me know and I'll add it.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/TokenEndpoint.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/TokenEndpoint.java
@@ -156,9 +156,9 @@ public class TokenEndpoint extends AbstractEndpoint {
 	}
 
 	@ExceptionHandler(HttpRequestMethodNotSupportedException.class)
-	public void handleHttpRequestMethodNotSupportedException(HttpRequestMethodNotSupportedException e) throws Exception {
+	public ResponseEntity<OAuth2Exception> handleHttpRequestMethodNotSupportedException(HttpRequestMethodNotSupportedException e) throws Exception {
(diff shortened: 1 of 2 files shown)


kelsin:
Hey guys, any thought on this? We'd love to get this fix into the next release if possible.

dsyer:
Probably makes sense. Did you fill out the contributor's agreement (link in README)?

kelsin:
Just signed, confirmation number: 120520150514030947

dsyer:
`HttpRequestMethodNotSupportedException` should be a 405 right? I'm quite surprised it isn't already. Maybe throwing an exception in an `@ExceptionHandler` is what stops that from happening.

kelsin:
You are right, let me change this to a 405 response instead of a 400.

But yes, throwing from the @ExceptionHandler puts this far above Spring's normal handling so it just comes out as a plain error out of the container, instead of being handled again.

kelsin:
Error is now a proper 405.

kelsin:
Let me know if there is anything else I can do on this to get it into a next release. Thanks!

dsyer:
Merged in 4a0574fde1

PR #475, opened 2015-05-08
Update DefaultTokenServices.java

gywangMtl:
add a getter so that the derived class don't need to have a var to store the TokenStore object.


dsyer:
I'm not a big fan of adding public methods unless there's a compelling need. What's the actual use case?

gywangMtl:
i guess the getter can be a protected.

the use case is when I extend the DefaultTokenServices, the sub class cannot access the tokenStore directly. I have to override the setter so the derived class can have a variable reference to the tokenStore object.

dsyer:
As there is a workaround and I don't know any valid reasons for extending `DefaultTokenServices` I think we can close this.

Issue #498, opened 2015-06-05
Token not stored after single sign on


KeesKoffeman:
After a successful login the restTemplate.getAccessToken() is called in order to store the token with a provided ClientTokenServices. See OAuth2ClientAuthenticationProcessingFilter.java line 117 +118.

		// Nearly a no-op, but if there is a ClientTokenServices then the token will now be stored
		restTemplate.getAccessToken();

The ClientTokenServices.saveAccessToken(OAuth2ProtectedResourceDetails resource, Authentication authentication, OAuth2AccessToken accessToken) is never called because the token is obtained directly from the scoped OAuth2ClientContext in OAuth2RestTemplate.

I'm using using version 2.0.7 with @EnableOAuth2Sso and a custom ClientTokenServices.

dsyer:
Isn't the problem that the `ClientTokenServices` is not present in the `OAuth2RestTemplate` (there isn't one by default if you are using only `@Enable*` annotations to set it up)?

KeesKoffeman:
It's present, I've added it as described in the manual.

dsyer:
I'm more or less convinced that the problem (if there is one) is not in Spring OAuth. I'll try and play with the sample SSO app in Spring Boot to convince myself, and hopefully others.

dsyer:
OK, changed my mind, I see the issue now: there is no `Authentication` when the token is acquired in the SSO filter, so it can't be saved in the `ClientTokenServices` until it is available in the security context.

Artgit:
When this feature will be available ? Also, is it safe to store token based only on authentication.getName(), resource.getClientId() and resource.getScope as implemented in DefaultClientKeyGenerator.extractKey method ?

I think there is a very good chance to face a collisions between OAuth2 providers in oauth_client_token table.

What do you think ?

ssozonoff:
Bump ?

At least please fix the documentation so its clear that this feature is not availble.
"http://projects.spring.io/spring-security-oauth/docs/oauth2.html" section "Persisting Tokens in a Client"

Thanks

Simbainashe:
Whats the status on this issue, I have implemented my custom ClientTokenServices in my client application to enable me to delete the tokens on logout but its not working. On checking the login transaction I noticed the ClientTokenServices code is not being executed.

PR #499, opened 2015-06-06
Save token after successful authentication when an ClientTokenServices is provided.

KeesKoffeman:
Fixes gh-498

In addition to this it would be nice if ClientTokenServices is injected by auto-configuration (even when null) to avoid customization of both the filter and the rest template.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/filter/OAuth2ClientAuthenticationProcessingFilter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/filter/OAuth2ClientAuthenticationProcessingFilter.java
@@ -16,20 +16,15 @@
 
 package org.springframework.security.oauth2.client.filter;
 
-import java.io.IOException;
-
-import javax.servlet.FilterChain;
-import javax.servlet.ServletException;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-
 import org.springframework.security.authentication.AuthenticationDetailsSource;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.authentication.BadCredentialsException;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.oauth2.client.OAuth2RestOperations;
 import org.springframework.security.oauth2.client.http.AccessTokenRequiredException;
+import org.springframework.security.oauth2.client.token.ClientTokenServices;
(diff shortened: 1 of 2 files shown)


dsyer:

KeesKoffeman:
I did right after I filed the pull request. I selected 'spring-security' and named you as project lead; is that correct?

dsyer:
That's fine. It's an automated system but I don't always get the notifications.

Artgit:
When this feature will be available ? Also, is it safe to store token based only on authentication.getName(), resource.getClientId() and resource.getScope as implemented in DefaultClientKeyGenerator.extractKey method ?

I think there is a very good chance to face a collisions between OAuth2 providers in oauth_client_token table.

What do you think ?

dsyer:
I think you have some very unusual providers if they have identical client id and scope names.

Artgit:
For example in my current implementation authentication.getName() returns null. This way I think I'll have the only one record (authentication_id will be based for many users only on the same client_id and same scopes) in  oauth_client_token table

dsyer:
Your `Authentication` is broken then, I guess. If you need to support null names then you should provide your own `ClientKeyGenerator`.

Artgit:
Yeah.. I see it now.. According to UserDetails.getUsername() method contract it can't return null.. Right now struggling with a following question http://stackoverflow.com/questions/34920743/spring-security-userdetails-and-username

pivotal-issuemaster:
@KeesKoffeman Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=499) to manually synchronize the status of this Pull Request.

kvandake:
Hello. What status it pull request? When will these changes committed?

pivotal-issuemaster:
@KeesKoffeman Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=499)!

bvulaj:
I've resyncd the PR and it looks like besides the conflict that this should be all set. Any chance of a merge soon?

jgrandja:
@bvulaj With the addition of `setClientTokenServices(ClientTokenServices)` on the `Filter`, this could only get into the next minor release which would be 2.4. We haven't scheduled 2.4 as of yet and it likely won't be released for a couple months at least.

jgrandja:
@KeesKoffeman As you may be aware, this project is scheduled to reach EOL. Please see the [announcement](https://spring.io/blog/2020/05/07/end-of-life-for-spring-security-oauth) for further details.

I'm going to close this PR as it doesn't make sense to merge these changes given the status of the project.

PR #559, opened 2015-08-25
Make sure scope parameter in redirect url is encoded

caspermout:
The authorization redirect url in the response location header can include spaces if there are multiple scopes. This is not valid and should be encoded.

This problem is introduced in version 2.0.7.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/filter/OAuth2ClientContextFilter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/filter/OAuth2ClientContextFilter.java
@@ -108,7 +108,7 @@ public class OAuth2ClientContextFilter implements Filter, InitializingBean {
 		}
 
 		this.redirectStrategy.sendRedirect(request, response, builder.build()
-				.toUriString());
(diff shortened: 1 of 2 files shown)


dsyer:
Thanks for the patch. Did you fill out the contributor's agreement (link in README)?

caspermout:
I just now filled out the contributor's agreement.

dsyer:
Thanks. I'm not 100% sure this is safe because we've had so many bug reports with encoding redirects. I assume all the tests pass?

caspermout:
All the test build passed on my machine and github has a little green thing saying:
All checks have passed

The behaviour was broken in release 2.0.7. Version 2.0.6 did correctly encode spaces (and probably other characters as well).
As it is now in version 2.0.7, it does not encode any url parameter values. Which is obviously not correct.

dsyer:
Any chance you could add a test then?

caspermout:
I did actually add a test named testTwoScopesRedirectUri in an existing test file called OAuth2ClientContextFilterTests.java.
If that is not enough, please be more specific.

dsyer:
Sorry, don't know how I missed that.

PR #566, opened 2015-09-01
Fixed javadoc problems that prevented javadoc generation

georgespalding:
The javadoc errors where preventing me from building the project, so I fixed them as best as I could.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -172,12 +172,6 @@
     <dependencies>
 		<dependency>
 			<groupId>org.springframework</groupId>
-			<artifactId>spring-aop</artifactId>
-			<version>${spring.version}</version>
-		</dependency>
-
-		<dependency>
-			<groupId>org.springframework</groupId>
 			<artifactId>spring-beans</artifactId>
 			<version>${spring.version}</version>
 		</dependency>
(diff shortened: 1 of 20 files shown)


dsyer:
Thanks for that. Some of those problems look like they would go away with Java 1.7, but we can probably just merge them anyway if you can confirm that you filled out the contributor's agreement (link in README).

georgespalding:
Hi all.
I’ve completed the form
https://support.springsource.com/spring_committer_signup

(I put Dave Syer as project lead)

Confirmation number 136420150902051356

====================================================================================
Another thing; the travis integration build does not agree with my pullrequest; Am I missing something?
See
https://travis-ci.org/spring-projects/spring-security-oauth/builds/78305168

/George

On 02 Sep 2015, at 10:21, Dave Syer <notifications@github.com<mailto:notifications@github.com>> wrote:

README

georgespalding:
Hi I've managed to reproduce the problem that occurs in the travis-ci build, I get it in about 1 out of 7-10 runs of the test org.springframework.security.oauth2.config.annotation.ResourceServerConfigurationTests.testWithAuthServerAndGlobalMethodSecurity

I'm pretty sure it's not releated to my changes, but to the order in which annotations/config is processed.
Running the test with debuglogging gives me interesting differences in what happens in the test.
A run that fails never creates a "shared instance of singleton bean 'org.springframework.security.oauth2.config.annotation.web.configuration.ResourceServerConfiguration'"
And all processing of ResourceServerConfiguration is skipped.
How could this be avoided? Perhaps I should file an issue.

dsyer:
Merged thanks

PR #580, opened 2015-09-24
Fixed Url

vivekkiran:
Fixed Url with the latest working one at http://www.eclipse.org/m2e/,  since the previous url was giving error.


dsyer:
Merged thanks.

PR #581, opened 2015-09-25
Fix for losing authentication details on token refresh

voronaam:
This fix is for the case when user implements custom
AccessTokenConverter that puts extra information into the token in
convertAccessToken and extracts it into Authentication inside
extractAuthentication.

Those details were lost on the refresh token code path.

The fix makes DefaultTokenServices to preserve the details much
like it is already doing so when user has supplied custom
AuthenticationManager.

The fix also changes createRefreshedAuthentication access level to
protected so that user can modify scope narrowing logic. That is
to conform to the access level of other similar methods, such as
isExpired.


dsyer:

voronaam:
I have signed it. Confirmation number: 148220151114050657

jgrandja:
@voronaam I'm going to close this PR in favour of #823 as per your [comment](https://github.com/spring-projects/spring-security-oauth/pull/823#issuecomment-242641675).

PR #608, opened 2015-10-24
If there is more information, the class will be OAuth2Exception, not the opposite.

cguZZman:
If we can get an OAuth2Exception already from the body, it is likely to have more information than the header does, so just re-throw it here.


cguZZman:
this pull request closes #618

PR #628, opened 2015-11-16
Travis build on JDK 6+7+8

praseodym:
Configures Travis to build on Oracle JDK 7+8 and OpenJDK 6+7.

Also enables migration to Travis' container-based infrastructure: http://docs.travis-ci.com/user/migrating-from-legacy/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: java
+jdk:
+  - oraclejdk8
+  - oraclejdk7
+  - openjdk7
+  - openjdk6
+sudo: false
 
 install: mvn -U install --quiet -DskipTests=true -P bootstrap
 script: mvn clean test -P bootstrap
-
-


praseodym:
Apparently build is broken on JDK 6:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project spring-oauth2-tests-common: Fatal error compiling: invalid target release: 1.7 -> [Help 1]
```

dsyer:
It should be fine to use the default JDK (which I think is openjdk8). Not sure why we get so many travis fails though.

praseodym:
Travis doesn't have OpenJDK 8: https://docs.travis-ci.com/user/languages/java#Testing-Against-Multiple-JDKs

dsyer:
Fine. My comment still stands (the default ought to be fine).

PR #629, opened 2015-11-17
Expire OAuth codes if they are not redeemed

yankee42:
RFC6749 [says](https://tools.ietf.org/html/rfc6749#section-4.1.2):

> The authorization code MUST expire shortly after it is issued to mitigate the risk of leaks.  A maximum authorization code lifetime of 10 minutes is RECOMMENDED.

However currently spring-security-oauth does not foresee any means of expiring authorization codes. This may in the long run pose a security risk at worst or a waste of hard disk/memory space at best.

The attached code will expire tokens by default after 10 minutes, but can be configured for any duration.

I have signed the contributor's agreement.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/code/InMemoryAuthorizationCodeServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/code/InMemoryAuthorizationCodeServices.java
@@ -1,5 +1,6 @@
 package org.springframework.security.oauth2.provider.code;
 
+import java.util.Iterator;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.springframework.security.oauth2.provider.OAuth2Authentication;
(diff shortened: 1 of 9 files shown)


kakawait:
Maybe not implement a blocking expire system? It may be a good candidate for executing clear in background on dedicated thread (scheduler)?

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/code/InMemoryAuthorizationCodeServices.java
```diff
@@ -11,18 +12,52 @@
...
+		private final long createdTimestamp;
+		private final OAuth2Authentication authentication;
+
+		private ExpiringOAuth2Authentication(final long createdTimestamp, final OAuth2Authentication authentication) {
+			this.createdTimestamp = createdTimestamp;
+			this.authentication = authentication;
+		}
+	}
+
+	public void setCodeLiftetimeSeconds(int codeLiftetimeSeconds) {
```
kakawait: If possible I prefer following signature `(long lifetime, TimeUnit lifetimeTimeUnit)` (or whatever for naming) but is less errors prone. Some example from other library (*Apache HttpClient*) https://github.com/apache/httpclient/blob/a287ab63878f5e5a47f9cde639f04bb9e370478e/httpclient5/src/main/java/org/apache/hc/client5/http/impl/sync/HttpClientBuilder.java#L803

pivotal-issuemaster:
@yankee42 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=629) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@yankee42 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=629)!

romanstrobl:
Hello, do you have any update on this PR? We would really like to use this functionality in our SW without implementing this feature outside of Spring Security.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/code/InMemoryAuthorizationCodeServices.java
```diff
@@ -11,18 +12,52 @@
...
 	@Override
 	public OAuth2Authentication remove(String code) {
-		OAuth2Authentication auth = this.authorizationCodeStore.remove(code);
-		return auth;
+		removeExpired();
+		ExpiringOAuth2Authentication expiringOAuth2Authentication = this.authorizationCodeStore.remove(code);
+		return expiringOAuth2Authentication == null ? null : expiringOAuth2Authentication.authentication;
+	}
+
+	private void removeExpired() {
```
jgrandja, line 39: I don't feel the expiring logic should be in this implementation. Instead, it should be in an external component that can be used for any `AuthorizationCodeServices` implementation. There is similar logic in `JdbcAuthorizationCodeServices` (in this PR) and `RedisAuthorizationCodeServices` would have to contain similar logic as well. As you can see, there is duplication of efforts across `AuthorizationCodeServices` implementations.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/code/JdbcAuthorizationCodeServices.java
```diff
@@ -66,6 +80,11 @@ public OAuth2Authentication mapRow(ResultSet rs, int rowNum)
 		return authentication;
 	}
 
+	private void removeExpired() {
```
jgrandja, line 83: Same comment applies as in `InMemoryAuthorizationCodeServices`

PR #649, opened 2015-12-03
Fix a typo in the oauth2 docs

tspaeth:
Just a very small fix. :-)

Side note: http://projects.spring.io/spring-security-oauth/docs/oauth2.html looks outdated - the link on "The OAuth2AuthenticationProcessingFilter is used to load the Authentication for the request given an authenticated access token" points to a wrong URL but in the github here it's already fixed.


dsyer:
Those things should be fixed now, thanks.

Issue #651, opened 2015-12-03
A code is generated even if the approval grant has status DENIED


gharizanov92:
Let's say you have an authorization server running on `http://localhost:8080` with an ApprovalStoreUserApprovalHandler with JdbcClientDetailStore and JdbcAprovalStore.

And let's say you make the following request:
`http://localhost:8080/oauth/authorize?grant_type=code&scope=sample_scope&esponse_type=code&amp;client_id=client&redirect_uri=https%3A%2F%2Fexample.com%2Fcb&secret=secret` (1)
Let's also say that the client with id client_id has only one scope `sample_scope`

So if you submit the approval form with the param `sample_scope` set to false (e.g. you decline). That would cause an access denied error and would forward you to 
`/cb?error=access_denied&error_description=User%20denied%20access`
Which is as expected. Problem is, if you hit the same URL (1) again, it would redirect you to the following sample URL:
`/cb?code=Nd7nmz`

As you can see, you get a valid code. Which is a big problem, because you can get an access token with this code.

If you take a look at ApprovalStoreUserApprovalHandler, method checkForPreApproval, line 143, you could see that each approval is added to validUserApprovedScopes without checking whether it was approved.  This should be in the if statement below on line 144. Because right now validUserApprovedScopes would always have the same elements as requestedScopes and the if statement on 156 will always return true. So if an approval was denied, it would still go to validUserApprovals and would case the autorizationRequest to have approved set to true.

I've created a pull request regarding this issue - `https://github.com/spring-projects/spring-security-oauth/pull/653`

PR #653, opened 2015-12-04
Fixed denied access status generating code

gharizanov92:
I've noticed that when you get a access_denied error, and you persist your approval grants to a DB, you would get a DENIED approval record. Without this change, all scopes are being added to validUserApprovalScopes which would cause authorizationRequest to be approved, even though all of the userApprovals could be with status DECLINED.

https://github.com/spring-projects/spring-security-oauth/issues/651


dsyer:

gharizanov92:
Hi, yes I just did. I've put Rob Winch under "project lead", hope this was correct.

PR #659, opened 2015-12-13
[jwt] support for setting type header

tony-kerz:
add support for setting type header stemming from https://github.com/cloudfoundry/uaa/issues/270


tony-kerz:
@dsyer i'm trying to get cloudfoundry/uaa#270 thru, and it is held up with this.

is there anything i can do to help expedite the process?

dsyer:
Thanks for the patch. I implemented it slightly differently to you but that change should be in 1.0.4.

dsyer:
1.0.4 is available now (probably still syncing to Maven Central).

Issue #655, opened 2015-12-08
Support for PKCE in server


dsyer:
If a client sends a `code_verifier` to the /authorize endpoint, then the /token endpoint should require it (in a different form) in the token request that acompanies the auth code. This is to mitigate attacks on mobile clients (which we don't support, but we do expect perhaps to be used with our server endpoints): https://www.rfc-editor.org/rfc/rfc7636.txt.

jgrandja:
Closing as per [comment](https://github.com/spring-projects/spring-security-oauth/pull/675#issuecomment-315128677)

PR #675, opened 2016-01-17
Support for PKCE in server as per OAUTH PKCE RFC

marcolenzo:
I implemented the OAUTH PKCE as per RFC 7636 https://www.rfc-editor.org/rfc/rfc7636.txt

I tried to be as less intrusive as possible on the existent code. I was wondering if I had to add the `code_challenge` and `code_challenge_method` parameters in the `AuthorizationRequest` but I eventually opted not to and to rely on the existent maps.

This should complete implementation for issue #655

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/exceptions/InvalidCodeVerifierException.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/exceptions/InvalidCodeVerifierException.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2006-2016 the original author or authors.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package org.springframework.security.oauth2.common.exceptions;
+
+/**
+ * @author Marco Lenzo
+ */
+@SuppressWarnings("serial")
+public class InvalidCodeVerifierException extends ClientAuthenticationException {
+
+	public InvalidCodeVerifierException(String msg, Throwable t) {
+		super(msg, t);
+	}
+
+	public InvalidCodeVerifierException(String msg) {
+		super(msg);
+	}
+
+	@Override
+	public String getOAuth2ErrorCode() {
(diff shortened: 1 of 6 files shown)


MrArtichaut:
Is there any plan to merge this into the master ? I've tried a local merge and there is no conflict and unit tests are all green.

The draft-ietf-oauth-native-apps now indicates that PKCE MUST be implemented for public native apps : https://tools.ietf.org/html/draft-ietf-oauth-native-apps-12#section-6 making this pull request even more interesting. 

Thanks.

osgafarov:
Should this PR be assigned to a specific person?
P.S. I am also working on a mobile app with PKCE approach, so merging this PR would be very helpful.

jgrandja:
Thanks for the PR @marcolenzo.
However, this project is currently in _maintenance mode_ and we're only accepting minor updates and bug fixes. We are not looking to add any new features.

We are focusing our efforts in the new OAuth2/OIDC support in Spring Security 5.

osgafarov:
@jgrandja Will there be PKCE support in Spring Security 5? If not, what would be the best way to authorize mobile apps to call REST API using OAuth2?

jgrandja:
@osgafarov Yes, we will provide support for PKCE in Spring Security 5, however, that will be a while still. Our current focus is in re-writing the client and that is what will go GA later this November. We're planning a 5.1 shortly afterward which will include the Resource Server support and then will start working on the Authorization Server side of things. So the Authorization Server support probably won't be out until mid 2018.

osgafarov:
Thanks @jgrandja for your answer. Will follow the development of the Spring Security 5.

raxim1:
I would suggest that instead of changing the code do it in a filter, like csrf() - so it would be pkce().
At least temporarily we can include without pain, spring security 5 is far away. Either way, everybody who heard about it tries to implement in his own way. At least i saw the code, so it's not too big deal to create some filter in the filterchain and add the implementation, which is independent so i don't need to hack the whole thing. The other thing is, that i'm using some sort of repository to store the client details, and i just don't see from the code changes, how it can be introduced.

OrangeDog:
@marcolenzo have you got an example (or advice) for implementing this without library changes (or minimal changes that can be applied with aop or replacing simple classes)?

marcolenzo:
@OrangeDog did you have the chance to revise the content of the PR itself? https://github.com/spring-projects/spring-security-oauth/pull/675/files 

An alternative option that does not fiddle with the internals of the library is to port the logic I introduced into a filter as suggested by @ghost https://github.com/spring-projects/spring-security-oauth/pull/675#issuecomment-321673694

OrangeDog:
Having a closer look this doesn't follow the RFC:

- Parameters should be checked on the auth requests, and "invalid_request" used as an error.
- If the verifier fails on the token request, the error is "invalid_grant".

The error "invalid_code_verifier" is not part of the standard.

OrangeDog:
You should probably also use constant-time comparisons, not `String.equals`.

PR #694, opened 2016-02-11
Fixes resetting the original auth object after chain.

rpvilao:
http://stackoverflow.com/questions/35345655/securitycontext-is-not-cleared-after-calling-the-chain-on-oauth2authenticationpr

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/authentication/OAuth2AuthenticationProcessingFilter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/authentication/OAuth2AuthenticationProcessingFilter.java
@@ -47,9 +47,9 @@ import org.springframework.util.Assert;
  * A pre-authentication filter for OAuth2 protected resources. Extracts an OAuth2 token from the incoming request and
  * uses it to populate the Spring Security context with an {@link OAuth2Authentication} (if used in conjunction with an
  * {@link OAuth2AuthenticationManager}).
- * 
+ *
  * @author Dave Syer
- * 
+ *
  */
 public class OAuth2AuthenticationProcessingFilter implements Filter, InitializingBean {
 
@@ -71,7 +71,7 @@ public class OAuth2AuthenticationProcessingFilter implements Filter, Initializin
 	 * Flag to say that this filter guards stateless resources (default true). Set this to true if the only way the
 	 * resource can be accessed is with a token. If false then an incoming cookie can populate the security context and
 	 * allow access to a caller that isn't an OAuth2 client.
-	 * 
+	 *
 	 * @param stateless the flag to set (default true)
 	 */
 	public void setStateless(boolean stateless) {
@@ -126,10 +126,12 @@ public class OAuth2AuthenticationProcessingFilter implements Filter, Initializin
 		final HttpServletRequest request = (HttpServletRequest) req;
 		final HttpServletResponse response = (HttpServletResponse) res;
 
+		final Authentication previousAuth = SecurityContextHolder.getContext().getAuthentication();
+
 		try {
 
 			Authentication authentication = tokenExtractor.extract(request);
-			
+
 			if (authentication == null) {
 				if (stateless && isAuthenticated()) {
 					if (debug) {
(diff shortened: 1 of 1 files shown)


jgrandja:
Hi @rpvilao. I've looked at your PR and your posting on SO and don't feel this is a bug. Are you able to provide me a sample that reproduces this issue that you are having? I'm going to close this for now but will re-open and investigate further if you can provide me a sample.

rpvilao-gymondo:
Hi,

The idea is the following: you'll need two endpoints one that requires auth and another that is public. You make the request to the secured one first and to the second after. The second request will have the authentication principal context of the first.

jgrandja:
@rpvilao-gymondo If the Bearer token is passed on the 2nd request in the Authorization header (just like the 1st request) then yes the `Authentication` will be the same on the 2nd request, as expected. Even though the 2nd request is a public endpoint with `permitAll` configured, the `OAuth2AuthenticationProcessingFilter` will still extract the `Authentication` from the incoming Bearer token...it just won't be used to enforce authorization given the 2nd request is configured with `permitAll`.

rpvilao-gymondo:
I know but it's happening without passing the token. It just takes the one from before because it's not cleared from the thread context.

jgrandja:
@rpvilao-gymondo I changed my test to not pass the Bearer Token on the 2nd request and the Authentication is **not loaded** from the 1st request. The `SecurityContext` is cleared after the 1st request completes as expected. As already mentioned, please provide a sample of your scenario that I can look at and help solve your issue. I cannot reproduce the same scenario you have described. Thanks.

rpvilao-gymondo:
Thanks for the effort you put into this. As soon as I have some spare time I'll get back to it. Thanks again!

rpvilao-gymondo:
So for me it's really a pain in the ass to provide you with a package to reproduce even because I'm using a database for the token store and I have separate servers for resource and oauth. The way I'm configuring it is the following:

```
    @Override
    public void configure(HttpSecurity http) throws Exception {
        final ExpressionUrlAuthorizationConfigurer<HttpSecurity>.ExpressionInterceptUrlRegistry registry =
                http.addFilterAfter(authorisationFilter(), SwitchUserFilter.class).authorizeRequests();

        HttpUtil.SECURED_METHODS.forEach(x -> registry.regexMatchers(x, SECURED_PATTERN).access(format("hasAnyRole('%s', '%s')", UserRole.USER, UserRole.LIMITED_USER)));
    }
```

Basically SECURED_METHODS is a list of methods that I configure to be under the secured realm. The secured pattern is just a regex: /([^/].*?)/secured/.*

The authorisation filter is a simple filter that only checks if the user's subscription is valid and it's not used for the public calls.

Then I have:
```
    @RequestMapping
    public Page<ProgramV1Dto> list(
            @AuthenticationPrincipal OAuth2UserDetails oAuth2UserDetails) {
        System.out.println(oAuth2UserDetails);
     }
```
And it prints:
**Without a token**
null 
**With a token**
OAuth2UserDetails{id=1, authorities=[USER], password=[PROTECTED], username=ola@example.com, accountNonExpired=true, accountNonLocked=true, credentialsNonExpired=true, enabled=true, client=androidtv, accessToken=null}
**Without the token again**
OAuth2UserDetails{id=1, authorities=[USER], password=[PROTECTED], username=ola@example.com, accountNonExpired=true, accountNonLocked=true, credentialsNonExpired=true, enabled=true, client=androidtv, accessToken=null}

jgrandja:
I really do need a sample in order to be effective in troubleshooting your issue. All I need is the Resource Server sample. For your `TokenStore`, you can configure an `InMemoryTokenStore` and set it on your `ResourceServerSecurityConfigurer`. Just make sure you save the `OAuth2AccessToken` into the `TokeStore` before you run your test.

rpvilao-gymondo:
Hi,

So I was isolating the code for you and I was testing and decided to write some scripts using curl and I realised that with curl I cannot reproduce it... only with postman via the browser. Meaning that the fact the browser is sending cookies (JSESSIONID probably) enable this to happen. Is there a way to prevent this?

jgrandja:
@rpvilao-gymondo No worries. I'm glad you got it working now.

PR #707, opened 2016-03-02
Fix thread-safety issue in DefaultOAuth2ClientContext

caspermout:
When using a RestTemplate created using <oauth2:rest-template and OAuth2ClientContextFilter it is possible for two threads(different requests) on the same web session to call  setPreservedState on the same DefaultOAuth2ClientContext object. This call uses HashMap.put which is not thread-safe.

This can result in all threads that handle requests for that session to enter an infinite loop with stacktrace:

```
"http-bio-8083-exec-47" daemon prio=10 tid=0x0000000002358000 nid=0x3b32 runnable [0x00007fef6b5f3000]
   java.lang.Thread.State: RUNNABLE
        at java.util.HashMap.put(HashMap.java:494)
        at org.springframework.security.oauth2.client.DefaultOAuth2ClientContext.setPreservedState(DefaultOAuth2ClientContext.java:53)
        at sun.reflect.GeneratedMethodAccessor250.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:606)
        at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:302)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:190)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:157)
        at org.springframework.aop.support.DelegatingIntroductionInterceptor.doProceed(DelegatingIntroductionInterceptor.java:133)
        at org.springframework.aop.support.DelegatingIntroductionInterceptor.invoke(DelegatingIntroductionInterceptor.java:121)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
        at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:207)
        at com.sun.proxy.$Proxy11.setPreservedState(Unknown Source)
        at org.springframework.security.oauth2.client.OAuth2RestTemplate.getAccessToken(OAuth2RestTemplate.java:184)
        at org.springframework.security.oauth2.client.OAuth2RestTemplate.createRequest(OAuth2RestTemplate.java:105)
        at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:587)
        at org.springframework.security.oauth2.client.OAuth2RestTemplate.doExecute(OAuth2RestTemplate.java:128)
        at org.springframework.web.client.RestTemplate.execute(RestTemplate.java:552)
        at org.springframework.web.client.RestTemplate.getForObject(RestTemplate.java:259)
        at sun.reflect.GeneratedMethodAccessor47.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:606)
```


caspermout:
I created a method to reproduce this issue, unfortunately it is not really suitable for in a test case. The test uses a thread local for the DefaultAccessTokenRequest object to simulate the request scope used by the <oauth2:rest-template/> tag and only one DefaultOAuth2ClientContext object to simulate the session scope.
On my machine the threads almost immediately enter the infinite loop. (I use jconsole to check this)
	
	private void testthreads() {
		
		final ThreadLocal<Object> ttt = new ThreadLocal<Object>();
		
		TargetSource accessTokenRequestTargetSource = new TargetSource() {
			
			@Override
			public Class<?> getTargetClass() {
				return Object.class;
			}

			@Override
			public boolean isStatic() {
				return false;
			}

			@Override
			public Object getTarget() throws Exception {
				if(ttt.get() == null) {
					ttt.set(new DefaultAccessTokenRequest());
				}
				return ttt.get();
			}

			@Override
			public void releaseTarget(Object target) throws Exception {
			}
		};
		
		final AccessTokenRequest accessTokenRequest = ProxyFactory.getProxy(AccessTokenRequest.class, accessTokenRequestTargetSource);
		
		OAuth2ClientContext context = new DefaultOAuth2ClientContext(accessTokenRequest);

		AuthorizationCodeResourceDetails resource = new AuthorizationCodeResourceDetails();
		
		final OAuth2RestTemplate restTemplate = new OAuth2RestTemplate(resource, context);
		
		try {
			restTemplate.getAccessToken();
		} catch(UserRedirectRequiredException e) {
			// ignore expected exception
		}
		
		final AtomicBoolean running = new AtomicBoolean(true);
		
		Thread[] threads = new Thread[10];
		for(int i = 0; i < threads.length; i++) {
			threads[i] = new Thread(new Runnable() {
				
				@Override
				public void run() {
					while(running.get()) {
						try {
							restTemplate.getAccessToken();
						} catch(UserRedirectRequiredException e) {
							// ignore expected exception
						}
						ttt.set(null);
					}
				}
			});
		}
		
		for(int i = 0; i < threads.length; i++) {
			threads[i].start();
		}
		
		try {
			Thread.sleep(300000);
		} catch (InterruptedException e) {
			e.printStackTrace();
		}
		
		running.set(false);
		
		for(int i = 0; i < threads.length; i++) {
			try {
				threads[i].join();
			} catch (InterruptedException e) {
				e.printStackTrace();
			}
		}
	}

dsyer:
The thread dump you pasted is not showing any kind of loop. Is there actually a bug in `HashMap`? I can't really tell because the line numbers don't match the JDK I am using. You are right that a `HashMap` is not thread safe, but it shouldn't hang (and it wouldn't really be a Spring problem if it did).

caspermout:
There are more stories on the internet about HashMap.put hanging when used from multiple threads at the same time, for example: http://stackoverflow.com/questions/22944918/why-does-the-code-hang-with-hashmap-put-from-multiple-threads 
But that does not really matter since the spring security oauth code should not use a non thread safe HashMap from multiple threads.

Is there a reason why ConcurrentHashMap cannot be used?

dsyer:
> Is there a reason why ConcurrentHashMap cannot be used?

No, I was just asking. Thanks for the link.

pivotal-issuemaster:
@caspermout Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=707) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@caspermout Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=707)!

jgrandja:
@caspermout I'm going to close this PR as it's quite old. If this is still valid, can you please rebase off master and resubmit a new PR.

PR #719, opened 2016-03-20
Return a more appropriate exception when extraction fails.

Fitzoh:
Currently in the `OAuth2ErrorHandler` if the extractor encounters an `HttpMessageConversionException`, that `HttpMessageConversionException` is thrown and no status code is propogated to the user.  This update catches and ignores `HttpMessageConversionException`'s  and proceeds with error handling, resulting in the user obtaining a more useful exception.


jgrandja:
Thanks for the PR @Fitzoh and for adding a test.
This has been merged to master.

Issue #722, opened 2016-03-23
More than 1 scope causes authentication to fail


michaelwoodson:
I had authentication working with 1 scope, then when I added an additional scope authentication started failing during access token extraction with a rather cryptic exception:

```
Caused by: com.fasterxml.jackson.databind.JsonMappingException: No content to map due to end-of-input
 at [Source: java.io.PushbackInputStream@88948e5; line: 1, column: 244]
	at com.fasterxml.jackson.databind.JsonMappingException.from(JsonMappingException.java:148) ~[jackson-databind-2.6.5.jar:2.6.5]
	at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:3781) ~[jackson-databind-2.6.5.jar:2.6.5]
	at com.fasterxml.jackson.databind.ObjectMapper._readValue(ObjectMapper.java:3694) ~[jackson-databind-2.6.5.jar:2.6.5]
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2005) ~[jackson-databind-2.6.5.jar:2.6.5]
	at com.fasterxml.jackson.core.JsonParser.readValueAs(JsonParser.java:1506) ~[jackson-core-2.6.5.jar:2.6.5]
	at org.springframework.security.oauth2.common.OAuth2AccessTokenJackson2Deserializer.deserialize(OAuth2AccessTokenJackson2Deserializer.java:87) ~[spring-security-oauth2-2.0.9.RELEASE.jar:na]
	at org.springframework.security.oauth2.common.OAuth2AccessTokenJackson2Deserializer.deserialize(OAuth2AccessTokenJackson2Deserializer.java:45) ~[spring-security-oauth2-2.0.9.RELEASE.jar:na]
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:3736) ~[jackson-databind-2.6.5.jar:2.6.5]
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2810) ~[jackson-databind-2.6.5.jar:2.6.5]
	at org.springframework.http.converter.json.AbstractJackson2HttpMessageConverter.readJavaType(AbstractJackson2HttpMessageConverter.java:222) ~[spring-web-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	... 68 common frames omitted
```

[...]

PR #728, opened 2016-03-27
Add support for scopes as JSON array

michaelwoodson:
Normally the scope is a space separated list, but at least one endpoint uses
a JSON array instead.  This change adds support for parsing a JSON array in
addition to a space separated list.

Fixes gh-722


jgrandja:
Thanks for the PR @michaelwoodson! This has now been merged to master.

PR #729, opened 2016-03-28
Updated broken link to OAuth2 docs

sthzg:
- changed link extension from `.md` to `.html`, as it is the case with the working `oauth1` link


dsyer:
Fixed independently, but thanks.

Issue #738, opened 2016-04-10
Consider expires_in: 0 in access token response as token without expiration


ivan-babiankou:
There are providers who return 0 for tokens without expiration.

Example: https://vk.com/dev/permissions 
For offline(token never expires) access documentation has comment that expires_in will have value 0.

rwinch:
Thanks for the report. We can make this an option to enable for the client

ivan-babiankou:
Sounds good to me. 
Just FYI I've submitted pull request with implementation shortly after created issue. 

Ivan

jgrandja:
Related to [gh-738](https://github.com/spring-projects/spring-security-oauth/pull/739)

jgrandja:
Hi @ivan-babiankou. Thanks for your feedback here. I've reviewed your PR, however, it doesn't provide a way to configure the default **expires_in** (`accessTokenValiditySeconds`) to 0 for newly generated access tokens. I see you made some updates to `OAuth2AccessTokenJackson1Deserializer` and `OAuth2AccessTokenJackson2Deserializer`.

However, if we were to provide a custom value for `accessTokenValiditySeconds` for new access tokens then this would need to be supplied to an implementation of `AuthorizationServerTokenServices`. The default implementation is `DefaultTokenServices` and it may be reused and all you would need to do is call `setAccessTokenValiditySeconds(0)` and this would generate non-expiring access tokens. For example, extend `AuthorizationServerConfigurerAdapter` and override the following method:

```
		@Override
		public void configure(AuthorizationServerEndpointsConfigurer endpoints) throws Exception {
			DefaultTokenServices tokenServices = new DefaultTokenServices();
			tokenServices.setAccessTokenValiditySeconds(0);
			endpoints.tokenServices(tokenServices);
		}

```

Additionally, there is another approach where you can configure individual clients with their own custom value for `accessTokenValiditySeconds`. For example, you can supply the `DefaultTokenServices` with an implementation of `ClientDetailsService` which ultimately returns a `ClientDetails` and has an accessor `getAccessTokenValiditySeconds`. This approach provides the most flexibility as you can configure a specific client to have non-expiring access tokens and another client to have an expiring access token.

```
	@Configuration
	@EnableWebSecurity
	@EnableAuthorizationServer
	public class AuthorizationServerTokenServicesConfig extends AuthorizationServerConfigurerAdapter {

		@Override
[...]

PR #739, opened 2016-04-10
#738 Consider access token with expires_in = 0 as never expiring

ivan-babiankou:
Straightforward implementation.


jgrandja:
Relates to [Issue 738](https://github.com/spring-projects/spring-security-oauth/issues/738)

PR #767, opened 2016-05-21
Allow XML configuration for allowed TokenEndpoint HTTP methods

michael-simons:
This is a response to #334. With this pull request, XML configuration can be used to configure the allowed request methods on the `TokenEndpoint` the same way it is possible with `AuthorizationServerConfigurerAdapter` without having to resort to a `BeanPostProcessor`:

```
<oauth:authorization-server token-endpoint-url="/token" token-endpoint-allowed-request-methods="GET, POST">
	<oauth:authorization-code />		
</oauth:authorization-server>
```

I'd really appreciate a merge, as it allows to keep an existing (and tested, working) XML configuration, especially when upgrading from 1.x to 2.x. Thanks!

I have signed and agree to the terms of the Spring Individual Contributor
License Agreement.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/xml/AuthorizationServerBeanDefinitionParser.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/xml/AuthorizationServerBeanDefinitionParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2008-2009 Web Cohesion
+ * Copyright 2008-2016 Web Cohesion
  * 
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
@@ -13,7 +13,10 @@
 
 package org.springframework.security.oauth2.config.xml;
 
+import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
+import java.util.Set;
 
 import org.springframework.beans.BeanMetadataElement;
 import org.springframework.beans.factory.config.RuntimeBeanReference;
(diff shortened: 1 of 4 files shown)


jgrandja:
Hi @michael-simons. Thanks for taking your time in putting together this PR but unfortunately we won't be able to accept this. It really goes against the OAuth spec to allow for a GET method request to obtain an access token.

As per spec, in Section 3.2. Token Endpoint

> The client MUST use the HTTP "POST" method when making access token requests.

I understand that this configuration has been exposed by `AuthorizationServerConfigurerAdapter`, which in hindsight, probably should not have been. I wouldn't advise the use of obtaining an access token using a GET request but if you absolutely have to you do have the workaround by using a `BeanPostProcessor`, as you are already aware of.

I'm going to close this PR. I hope to hear from you again soon.

michael-simons:
Hi @jgrandja 
Too bad. I referred to the Spec and I am aware of it but I'd prefer a consistent configuration behavior, even if the Spec is absolutely right (with the password appearing in logs etc. for example). 
Anyway, no offences taken. 
Michael.

jgrandja:
@michael-simons Yes consistency is key for sure. But we don't want to open the door more than it can at this point. We'll strive to be consistent going forward.

shivaprasad573:
I have added the code that you are proposed but getting error in xml file .. do i need to change any schema.

**cvc-complex-type.3.2.2: Attribute 'token-endpoint-allowed-request-methods' is not allowed to 
 appear in element 'oauth:authorization-server'.**

My spring-security.xml file is
<beans xmlns="http://www.springframework.org/schema/beans"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:oauth="http://www.springframework.org/schema/security/oauth2"
	xmlns:context="http://www.springframework.org/schema/context"
	xmlns:sec="http://www.springframework.org/schema/security" xmlns:mvc="http://www.springframework.org/schema/mvc"
	xsi:schemaLocation="http://www.springframework.org/schema/security/oauth2 http://www.springframework.org/schema/security/spring-security-oauth2.xsd
		http://www.springframework.org/schema/mvc http://www.springframework.org/schema/mvc/spring-mvc-3.2.xsd
		http://www.springframework.org/schema/security http://www.springframework.org/schema/security/spring-security-3.2.xsd 
		http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-4.1.xsd
		http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-4.1.xsd ">

PR #823, opened 2016-08-16
Preserve user authentication details on re-authentication

klieber:
During a token refresh in the `DefaultTokenServices` the user
authentication will be re-authenticated if an `AuthenticationManager`
was provided. A `PreAuthenticatedAuthenticationToken` is created
based on the user authentication and then passed to the
`AuthenticationManager`. However, if there were any details on
the user authentication those details are lost because they are
not copied to the `PreAuthenticatedAuthenticationToken`. If the
`AuthenticationManager` is not provided then this logic is skipped
over and the details are correctly preserved.

The fix is simply to set the details on the
`PreAuthenticatedAuthenticationToken` before passing it to the
`AuthenticationManager`.  I've moved the re-authentication code into
a new protected method named `reauthenticateUserAuthentication` so that
it can be overridden if needed.

Finally, I added two new tests to `DefaultTokenServicesTests` to
validate that the user authentication is built correctly on a refresh.
There is one test for the scenario when there is no re-authentication
which passes even before these changes and then the other tests the
re-authentication scenario which requires this change to pass.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
@@ -155,8 +155,16 @@ public class DefaultTokenServices implements AuthorizationServerTokenServices, R
 		if (this.authenticationManager != null && !authentication.isClientOnly()) {
 			// The client has already been authenticated, but the user authentication might be old now, so give it a
 			// chance to re-authenticate.
-			Authentication user = new PreAuthenticatedAuthenticationToken(authentication.getUserAuthentication(), "", authentication.getAuthorities());
-			user = authenticationManager.authenticate(user);
+			Authentication userAuthentication = authentication.getUserAuthentication();
+			PreAuthenticatedAuthenticationToken preAuthenticatedToken = new PreAuthenticatedAuthenticationToken(
+					userAuthentication,
+					"",
+					authentication.getAuthorities()
+			);
+			if (userAuthentication != null) {
+				preAuthenticatedToken.setDetails(userAuthentication.getDetails());
+			}
+			Authentication user = authenticationManager.authenticate(preAuthenticatedToken);
 			Object details = authentication.getDetails();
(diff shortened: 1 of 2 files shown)


klieber:
This change is loosely related to #581, however, the fix in that pull request does not fix the issue I was having.  I believe that is because my issue is with the details on the user authentication whereas the @voronaam was concerned with the details on the `OAuth2Authentication` object.

klieber:
I have signed the contributors agreement.

voronaam:
Looks like your change will fix my problem as well. I hope it will get accepted. If it does, I'll retest my specific use case and will reject my PR if it is no longer needed.
Thank you!

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
```diff
@@ -211,6 +207,32 @@ private OAuth2Authentication createRefreshedAuthentication(OAuth2Authentication
 		return narrowed;
 	}
 
+	/**
+	 * Re-authenticate the user authentication to verify it is up-to-date.
+	 * @param authentication The authentication
+	 * @return The re-authenticated authentication.
+	 */
+	protected OAuth2Authentication reauthenticateUserAuthentication(OAuth2Authentication authentication) {
```
jgrandja, line 215: Please remove `reauthenticateUserAuthentication()` and `createUserPreAuthenticatedToken()` as it's not needed. The only change required is to `setDetails()` on the `PreAuthenticatedAuthenticationToken` before it's passed to `AuthenticationManager`.

klieber: > Thanks for the PR @klieber. If you're still looking to get this merged, please see my review comment and apply the change and we'll get this merged.

Sure I will take a look.

klieber: @jgrandja I've made the changes requested.  Let me know if there is anything else needed.

jgrandja:
Thanks for the update @klieber. This is now in master and backported to 2.4.x.

Issue #808, opened 2016-07-16
User can get token via client_credentials or password grants


socket70:
This is a simple authorization server based on the samples. The only real difference is I plug in a custom UserDetailsService and ClientDetailsService. They just hardcode the credentials but I was working on implementing the database when I came across this.

The problem: if you have both a client and a user with the same ID (clientId and username) the user can get an access token using the client_credentials or password grant flow.

[Here is the sample code](https://github.com/socket70/SpringOAuth2AuthenticationProvider).

Am I doing something wrong or is this a bug?

Thanks.

benkiefer:
@dsyer This is a security issue that is easily reproducible. Can you take a look at this?

benkiefer:
@jgrandja saw you made several commits to this repository, and wanted to see if you could look at this?

We were able to patch it in our own code base by overriding the appropriate endpoints and adding pre-checks to ensure the authorization is the right type/user.

This exploit is easy to reproduce because the client id is part of the query parameters for the authorization flow.

jgrandja:
@benkiefer @socket70 I'm looking into this now.

jgrandja:
@socket70 @benkiefer Thank you for reporting this! A fix has been applied and has been merged to master. Release is going out shortly.

benkiefer:
@jgrandja Thank you!

jgrandja:
See [blog post](https://spring.io/blog/2016/08/31/spring-security-oauth2-client-authentication-issue)

socket70:
Thank you @jgrandja

benkiefer:
Thank you!

jfneis:
Please ignore my previous questions. After rebuilding the whole application I found a type in the check_token endpoint, what was causing Spring to consider it as a regular endpoint (as expected!).

Miguelme:
I'm getting a similar issue where Spring is trying to load the client credential grant from the userDetailsService as if it were a password grant type instead of client credentials. Do you have any ideas?

jfneis:
@Miguelme not really, but our problem was related to trying to access a wrong check_token URL (so Spring would correctly think it was a regular access and try to validate the token).

Using oauth/check_token and sending the "token" header worked perfectly.

PR #824, opened 2016-08-18
Prevent user from obtaining token via client_credentials or password grant

jgrandja:
Fixes gh-808


dsyer:
LGTM

rwinch:
Can we add a test for the XML use case?

rwinch:
We should also provide a link to a sample that demonstrates the app broken and then fixed for both Java Config and XML config. The link should be posted on the issue and then also posted on our blog post with the release. The primary reason for this is for the XML configuration we cannot just say update to the latest. We need to ensure users know how to configure it correctly.

jgrandja:
XML tests have been added for valid and invalid configuration.

jgrandja:
@rwinch Are you ok if we provide a link to the Java Config and Xml Config Tests in the blog post and Issue instead of the sample repo you requested? The tests demonstrate a valid and invalid Xml Config and as far as the Java Config goes all the user needs to do is update to 2.0.11.RELEASE.
@dsyer and I spoke and we feel this is sufficient. What do you think?

rwinch:
That seems ok to me

Issue #717, opened 2016-03-16
Forced header Basic Authorization  in RemoteTokenServices.class


wojt12322:
Hi,
When i setUp  in configuration class @EnableAuthorizationServer

```
@Override
    public void configure(final AuthorizationServerSecurityConfigurer security) throws Exception {
        security.checkTokenAccess("permitAll()"); 
    }
```
My authorization server allows anyone to check tokens, without authentication. 

When  in @EnableResourceServer adds checkTokenEndpoint without credentials
```
    @Bean
    public RemoteTokenServices remoteTokenServices() {
        final RemoteTokenServices remoteTokenServices = new RemoteTokenServices();
        remoteTokenServices.setCheckTokenEndpointUrl("http://localhost:9999/authorization-server/oauth/check_token");

        return remoteTokenServices;
    }
```

RemoteTokenServices.class allways added Basic Authroization Header

```
 @Override
	public OAuth2Authentication loadAuthentication(String accessToken) throws AuthenticationException, InvalidTokenException {

		MultiValueMap<String, String> formData = new LinkedMultiValueMap<String, String>();
		formData.add(tokenName, accessToken);
		HttpHeaders headers = new HttpHeaders();
		headers.set("Authorization", getAuthorizationHeader(clientId, clientSecret));
		Map<String, Object> map = postForMap(checkTokenEndpointUrl, formData, headers);

		if (map.containsKey("error")) {
			logger.debug("check_token returned error: " + map.get("error"));
			throw new InvalidTokenException(accessToken);
		}

		Assert.state(map.containsKey("client_id"), "Client id must be present in response from auth server");
		return tokenConverter.extractAuthentication(map);
	}

	private String getAuthorizationHeader(String clientId, String clientSecret) {
		String creds = String.format("%s:%s", clientId, clientSecret);
		try {
			return "Basic " + new String(Base64.encode(creds.getBytes("UTF-8")));
		}
		catch (UnsupportedEncodingException e) {
			throw new IllegalStateException("Could not convert String");
		}
	}

```

In BasicAuthenticationFilter.class 
When Authorization header is a Basic, is followed checking users.

```
	protected void doFilterInternal(HttpServletRequest request,
			HttpServletResponse response, FilterChain chain) throws IOException,
			ServletException {
		final boolean debug = logger.isDebugEnabled();

		String header = request.getHeader("Authorization");

		if (header == null || !header.startsWith("Basic ")) {
			chain.doFilter(request, response);
			return;
[...]

PR #829, opened 2016-08-24
Added warning message to the RemoteTokenServices when attempting to u…

ChrisCooney:
…se a null client ID or client secret. Related to question raised in https://github.com/spring-projects/spring-security-oauth/issues/717

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
@@ -123,7 +123,7 @@ public class RemoteTokenServices implements ResourceServerTokenServices {
 	private String getAuthorizationHeader(String clientId, String clientSecret) {
 
 		if(clientId == null || clientSecret == null) {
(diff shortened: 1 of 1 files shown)


spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
```diff
@@ -121,6 +121,11 @@ public OAuth2AccessToken readAccessToken(String accessToken) {
 	}
 
 	private String getAuthorizationHeader(String clientId, String clientSecret) {
+
+		if(clientId == null || clientSecret == null) {
+			logger.debug("Null Client ID or Client Secret detected. Endpoint that requires authentication will reject request with 401 error.");
```
dsyer: DEBUG is a bit of a weak signal, don't you think, given that this is almost certainly going to fail?

ChrisCooney: Fair point. Maybe logger.warn might be a little more sensible.

PR #830, opened 2016-08-24
OAuth2 MongoDB default implementation

marcosbarbero:
This PR adds a default implementation for MongoDB.

**Note**: I've changed `${spring.version}` from `4.0.9.RELEASE` to `4.1.8.RELEASE`, if it's not allowed let me know. It was done because of `spring-data-mongodb` version requirement.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <commons-codec.version>1.9</commons-codec.version>
-    <spring.version>4.0.9.RELEASE</spring.version>
+    <spring.version>4.1.8.RELEASE</spring.version>
     <spring.security.version>3.2.8.RELEASE</spring.security.version>
     <java.version>1.6</java.version>
   </properties>
--- a/spring-security-oauth2/pom.xml
+++ b/spring-security-oauth2/pom.xml
@@ -16,6 +16,8 @@
 		<jackson1.version>1.9.13</jackson1.version>
 		<jackson2.version>2.3.2</jackson2.version>
 		<spring.security.jwt.version>1.0.2.RELEASE</spring.security.jwt.version>
+		<spring-data-mongodb.version>1.8.4.RELEASE</spring-data-mongodb.version>
+		<de.flapdoodle.embed.mongo.version>1.50.5</de.flapdoodle.embed.mongo.version>
 	</properties>
 
 	<build>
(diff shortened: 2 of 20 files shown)


RichardCSantana:
:+1:

cschwan2000:
👍

cschwan2000:
thx @jgrandja

jgrandja:
Hi @marcosbarbero. Thank you for your efforts in this PR.

Unfortunately, this project is in maintenance mode going forward and are limiting new features and only resolving bugs or minor enhancements.

FYI, we are re-writing OAuth and integrating it into the Spring Security core. You can track the issue [here](https://github.com/spring-projects/spring-security/issues/3907).

I did spend some time in reviewing your implementation and we may consider this feature in the new OAuth implementation depending on demand. But I did want to point out a couple of things.

Our preference would be to include this feature and it's dependencies in a module that may be explicitly included by the user if required. Having the mongo dependencies in the main POM is not ideal. I know redis is included there now but things can and will get out of hand if we start including more data stores in the main POM. Module is the preference for this type of extended feature.

Also, I would be very reluctant to include the test dependency `de.flapdoodle.embed.mongo`. I took a look at this project briefly and it's fairly new with little activity thus far. We would only include 3rd party dependencies that have been proven to be stable and is widely supported.

Please do keep track of the new OAuth initiative within Spring Security and we hope you can contribute there. First milestone will be coming out real soon in March.

marcosbarbero:
@jgrandja thanks for sharing all this info I really appreciate that.

As you already know somethings was done in that way because of how the project is written today e.g. the mongo dependencies. The decision to use `de.flapdoodle.embed.mongo` was based on [Spring Boot Embedded MongoDB dependencies](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-dependencies/pom.xml#L801) of course I can look for another approach but as long as it is being used for other spring projects I thought it could be a good option and even better than have an external dependency (standalone mongo running) to run the tests, just like it happens on Redis, at least when I wrote this PR there was a need of a up and running Redis on localhost.

Anyway I'll keep an eye on this new version of spring security and hopefully to contribute for it.

jgrandja:
Thanks @marcosbarbero for pointing out that `de.flapdoodle.embed.mongo` is being used in `spring-boot-dependencies`. Makes sense that you would use this based on the fact Spring Boot is using it.

Either way, the module approach is preferred but given the state of way things are organized today we are limited.

I'm going to close this PR but please keep an eye out on the new OAuth coming out soon.

marcosbarbero:
Thanks @jgrandja I'll keep an eye on the new approach.

Issue #836, opened 2016-08-27
Parsing scopes that are space-delimited rather than JSON arrays


mekangas:
The resolution of #745 supports a scope value consisting of a single logical string.  However, authorization servers (e.g., [OpenAM)](https://forgerock.org/openam/) whose "check token" endpoint implements the [draft OAuth 2.0 Token Introspection](https://tools.ietf.org/html/draft-ietf-oauth-introspection-04) standard returns scopes as a space-delimited string rather than a JSON array. `DefaultAccessTokenConverter` does not correctly support such scope values (`extractAuthentication` incidentally functions correctly following the resolution of #745 by virtue of `BaseRequest`, but `extractAccessToken` does not).

Issue #745, opened 2016-05-03
Allow single "scope" in JWT without need for JSON array format?


vladimirlapacek:
Similarly to #557, some authorisation servers are producing JWTs with single scope, which is not enclosed in an array. IdentityServer3 using .Net JwtSecurityTokenHandler may be an example. Such tokens make `DefaultAccessTokenConverter.extractAuthentication` and `DefaultAccessTokenConverter. extractAccessToken` throw an exception because it expects only an array. Can we make the `DefaultAccessTokenConverter` to cope with a single scope also?

jgrandja:
Thanks for the report @vladimirlapacek. This support has been added and is now in master.

PR #839, opened 2016-09-02
Support for space-delimited scopes in DefaultAccessTokenConverter

mekangas:
Fixed `DefaultAccessTokenConverter` to correctly handle, in all cases, the extraction of multiple scopes expressed as a space-delimited string rather than a JSON array in resolution to #836.  E.g., the [draft OAuth 2.0 Token Introspection](https://tools.ietf.org/html/draft-ietf-oauth-introspection-11#section-2.2) standard specifies such scopes.  (`DefaultAccessTokenConverter.extractAuthentication()` incidentally works for such scopes due to side effects from the resolution of #745; however, `extractAccessToken()` does not.)


jgrandja:
Thanks for the PR @mekangas.
This has now been merged to master.

PR #840, opened 2016-09-02
Removed client_id check in RemoteTokenServices

### LINKED ISSUE #838: RemoteTokenServices.loadAuthentication() should not check for client_id
Per [RFC 6749](https://tools.ietf.org/html/rfc6749#section-7) (which doesn't address the topic) and [RFC 7662](https://tools.ietf.org/html/rfc7662#section-2.2) (which explicitly permits this), a valid _check token_ response from the authorization server might not contain a `client_id`.  Therefore, it isn't appropriate for `RemoteTokenServices.loadAuthentication()` to validate this.  Rather, the contents of a token from an otherwise valid (i.e., non-error) response is the business of, say, `OAuth2AuthenticationManager.authenticate()`.  The check currently in place limits the usefulness of `RemoteTokenServices`, especially for bearer tokens.

mekangas:
Fixes #838 by removing the check for `client_id` in `RemoteTokenServices.loadAuthentication()`.  This check is inappropriate as it's the `ResourceServerTokenServices`' responsibility only to load tokens, not to judge them (other than as invalid in the event of an explicit error response). [RFC 6749](https://tools.ietf.org/html/rfc6749#section-7) doesn't specify the contents of such _check token_ endpoint responses while, for example, [RFC 7662](https://tools.ietf.org/html/rfc7662#section-2.2) explicitly makes `client_id` optional for `introspection` endpoints.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
@@ -111,7 +111,6 @@ public class RemoteTokenServices implements ResourceServerTokenServices {
 			throw new InvalidTokenException(accessToken);
 		}
 
-		Assert.state(map.containsKey("client_id"), "Client id must be present in response from auth server");
 		return tokenConverter.extractAuthentication(map);
 	}
 


mekangas:
Fixes #838

jgrandja:
Closing this PR in favour of f1a9b97.
Please see [comment](https://github.com/spring-projects/spring-security-oauth/issues/838#issuecomment-301554015)

PR #859, opened 2016-09-29
fix NPE while recreating token in case of existing non expiring token

igrik512:
Hi all!
Thanks for your awesome framework.
I think i've found a bug.
When we have a non-exipiring token and client application wants to re-create token, sending to oauth/autorize path with parameters. Then we will get NPE in method 

```
ApprovalStoreUserApprovalHandler.checkForPreApproval(AuthorizationRequest authorizationRequest,
            Authentication userAuthentication)
```

at this line 

```
if (approval.getExpiresAt().after(today))
```

So the rootcause is :
We configuring clients with ClientDetailsServiceConfigurer setting accessTokenValiditySeconds to 0. (So our token should never be expired)

In this case when the token will be created at method 

```
DefaultTokenServices.createAccessToken(OAuth2Authentication authentication, OAuth2RefreshToken refreshToken)
```

we have a condition 

```
if (validitySeconds > 0) {
            token.setExpiration(new Date(System.currentTimeMillis() + (validitySeconds * 1000L)));
        }
```

And Approval will be created (in method 

```
TokenApprovalStore.getApprovals(String userId, String clientId)
```

) with null expireAt Date.
That's why NPE appears.

I've attached a solution for null checking and tests for all usages, that i've found.
Maybe it has more elegant solution. Please let me know.
Thanks again for your work:)

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/approval/ApprovalStoreUserApprovalHandler.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/approval/ApprovalStoreUserApprovalHandler.java
@@ -139,7 +139,7 @@ public class ApprovalStoreUserApprovalHandler implements UserApprovalHandler, In
 		// Look at the scopes and see if they have expired
 		Date today = new Date();
 		for (Approval approval : userApprovals) {
-			if (approval.getExpiresAt().after(today)) {
+			if (approval.getExpiresAt() == null || approval.getExpiresAt().after(today)) {
 				if (approval.getStatus() == ApprovalStatus.APPROVED) {
 					validUserApprovedScopes.add(approval.getScope());
 					approvedScopes.add(approval.getScope());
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@igrik512 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=859)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=859) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@igrik512 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=859)!

jgrandja:
@igrik512 I'm going to close this PR as it's quite old. If this is still valid, can you please rebase off master and resubmit a new PR.

Issue #853, opened 2016-09-27
oauth_access_token Duplicate entry for PRIMARY key, MySQLIntegrityConstraintViolationException


maxim-filkov:
Hello,

I'm experiencing a problem with spring-security-oauth (2.0.11.RELEASE) that is trying to create duplicating access tokens. On attempt to get access_token by refresh_token:
```
curl -X POST -vu app123:TBe6G8W3t7Uv http://localhost:8090/oauth/token -H "Accept: application/json" -d "refresh_token=eyJhbGciOiJIUz...Wk_vs2nSk&grant_type=refresh_token&scope=read%20write&client_secret=TBe6G8W3t7Uv&client_id=app123"
```
I get:
```
{"error":"server_error","error_description":"PreparedStatementCallback; SQL [insert into oauth_access_token (token_id, token, authentication_id, user_name, client_id, authentication, refresh_token) values (?, ?, ?, ?, ?, ?, ?)]; Duplicate entry '0bc2efb38e22bccb7e6f1f565165d243' for key 'PRIMARY'; nested exception is com.mysql.jdbc.exceptions.jdbc4.MySQLIntegrityConstraintViolationException: Duplicate entry '0bc2efb38e22bccb7e6f1f565165d243' for key 'PRIMARY'"}
```
Any help would be really appreciated.

**Database table oauth_access_token**
```
`oauth_access_token` (
  `token_id` varchar(256) DEFAULT NULL,
  `token` blob,
  `authentication_id` varchar(256) NOT NULL,
  `user_name` varchar(256) DEFAULT NULL,
  `client_id` varchar(256) DEFAULT NULL,
  `authentication` blob,
  `refresh_token` varchar(256) DEFAULT NULL,
  PRIMARY KEY (`authentication_id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```
**Database config**
```
@Configuration
@EnableTransactionManagement
public class MySQLConfig {

   @Value("#{environment.MARIADB_PORT_3306_TCP_ADDR}")
   private String serverMySQLHost;

   @Value("#{environment.MARIADB_PORT_3306_TCP_PORT}")
   private String serverMySQLPort;

   @Value("#{environment.MARIADB_ENV_MYSQL_ROOT_PASSWORD}")
   private String serverMySQLRootPassword;

   /**
    * Data source for MySQL database (mainly used to save user credentials).
    *
    * @return Data source bean.
    */
   @Bean
[...]

PR #866, opened 2016-10-04
Fix for duplicated access tokens

maxim-filkov:
There is an attempt to create duplicated access tokens in oauth_access_token database table (attempt to insert rows with same token_id). This is related to the following [question](https://github.com/spring-projects/spring-security-oauth/issues/853).

[Preconditions]
1. JdbcTokenStore is used to persist the tokens.
2. JwtAccessTokenConverter is used to enhance the tokens.
3. oauth_access_token and oauth_refresh_token are empty.

[Steps]
1. Request access token by login/password pair.
```
curl -X POST -vu app123:TBe6G9W1e8Uv http://localhost:8090/oauth/token -H "Accept: application/json" -d "password=qwerty&username=max&grant_type=password&scope=read%20write&client_secret=TBe6G9W1t7Uv&client_id=app123"
```
Authorization service returns access token with refresh token.
```
{"access_token":"eyJasdD...SsdDss","token_type":"bearer","refresh_token":"eysdSJ...bnSDsc","expires_in":59,"scope":"read write","jti":"486f14dd-eaea3-422e-9f50-852072ba8b7e"}
```
2. Refresh access token by previously received refresh token.
```
curl -X POST -vu app123:TBe6G9W1t7Uv http://localhost:8090/oauth/token -H "Accept: application/json" -d "refresh_token=eysdSJ...bnSDsc&grant_type=refresh_token&scope=read%20write&client_secret=TBe6G9W1t7Uv&client_id=app123"
```
The existing access token is removed. In DefaultTokenServices.refreshAccessToken when creating new access token the given refresh token is overwritten with a new one (when enhancing it) here:
[JwtAccessTokenConverter.java](https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/JwtAccessTokenConverter.java#L15)
```
line 236: DefaultOAuth2RefreshToken token = new DefaultOAuth2RefreshToken(encode(
					encodedRefreshToken, authentication)); 
```
Then this is saved into oauth_access_token table. So oauth_access_token and oauth_refresh_token become out of sync (oauth_refresh_token has the previous refresh token and oauth_access_token (refresh_token column) has the new generated one).
3. Refresh access token once again by the same refresh token received in step 1.
Because we can't find the access token by the given refresh token, a duplicated access token is inserted into oauth_access_token table (with same token_id).

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
@@ -173,11 +173,14 @@ public class DefaultTokenServices implements AuthorizationServerTokenServices, R
 			refreshToken = createRefreshToken(authentication);
 		}
 
-		OAuth2AccessToken accessToken = createAccessToken(authentication, refreshToken);
-		tokenStore.storeAccessToken(accessToken, authentication);
+		DefaultOAuth2AccessToken accessToken = new DefaultOAuth2AccessToken(
+				createAccessToken(authentication, refreshToken));
 		if (!reuseRefreshToken) {
 			tokenStore.storeRefreshToken(accessToken.getRefreshToken(), authentication);
+		} else {
+			accessToken.setRefreshToken(refreshToken);
 		}
+		tokenStore.storeAccessToken(accessToken, authentication);
 		return accessToken;
 	}
 


jgrandja:
Hi @maxim-filkov. Thank you for submitting this PR.

There are some failing tests that may be a result of your changes? Can you please ensure all tests are passing.

Also, can you please add new tests for your proposed fix.

Thank you!

maxim-filkov:
Hi Joe, this PR actually doesn't fix the problem but rather can be considered as a workaround or an extended explanation what the issue is. Fixing this issue from my point of view implies significant code changes. The problem is that when request refresh for access token by refresh token, the refresh token (valid one) is overwritten inside JwtAccessTokenConverter (enhance method), here: 
```
DefaultOAuth2RefreshToken token = new DefaultOAuth2RefreshToken(encode(
					encodedRefreshToken, authentication));
```
The this overwritten refresh token is saved in oauth_access_token database table (refresh_token column) but never updated in the oauth_refresh_token database table. On next request to refresh access token by the same (still valid) refresh token we can't find access token by the given refresh token in the table oauth_access_token (because it was removed on the previous step). So we try to insert a duplicated access token (with same token_id).

My idea is that the enhance method should overwrite or reuse the given refresh token based on the "reuseRefreshToken" (true/false) setting. But the method doesn't consider this.

Because the TokenEnhancer interface (which is implemented by JwtAccessTokenConverter) is used in different places across the project, I can't just easily change the signature of the "enhance" method without breaking a lot of things.

I'm sorry for such a verbose weird explanation, but I do need some help here for fixing the defect.

jgrandja:
Thanks for all the input @maxim-filkov.
I'll attempt to reproduce the issue and go from there.

jgrandja:
Hi @maxim-filkov. Thanks again for the detailed info as I was able to reproduce the issue.

This issue is actually more involved to fix. Let me try to explain in detail.

Based on your example, you initiate a `password` grant flow and upon success you receive an `AccessToken` and `RefreshToken`.

Your JWT `RefreshToken` looks similar to this:

```
{
  "exp": 1478888786,
  "user_name": "user1",
  "jti": "5a28c182-7278-41ec-8b64-dfb6b757f8ed",
  "client_id": "client1",
  "scope": [
    "read",
    "write"
  ],
  "ati": "9e101185-f128-4fa1-a8f7-33cb88770cf9"
}
```

The key part in that JWT to note is:

```
"ati": "9e101185-f128-4fa1-a8f7-33cb88770cf9"
```

this is actually the `token id` (or reference) to your `AccessToken`.

Now, the next step you take is a `refresh_token` grant flow passing the `RefreshToken`. What happens next in `DefaultTokenServices` is a new `AccessToken` is created and it's ultimately **Enhanced** through `JwtAccessTokenConverter`. The `AccessToken` and `RefreshToken` get enhanced. 

The important part here is that the enhanced `RefreshToken` now contains a different `token id` reference to the newly created `AccessToken`, for example:

```
"ati": "*** reference to new Access Token Id ***"
```

**This enhancement effectively invalidates the original `RefreshToken`**. This is why you are getting an error on the 2nd (and any subsequent) `refresh_token` grant flows.

With the current design using JWT tokens, the `RefreshToken` can only be used once. Obviously not ideal at all. This will take a little more effort to fix than I was hoping.

As a quick workaround, each time you refresh an `AccessToken`, ensure you use the returned `RefreshToken` on the next `refresh_token` grant flow. You will also need to ensure that `reuseRefreshToken = false` for `DefaultTokenServices` to make this work.

Please give this workaround a try and let me know if it works on your end. I have a test on my end that works.

In the meantime, let me see what I can come up with to fix this bigger issue.

jgrandja:
@dsyer Do you have any suggestions on fixing this issue? 

Given that a JWT `RefreshToken` contains a reference to the `AccessToken` ID through the `ati` attribute, it ultimately gets invalidated upon an `AccessToken` refresh because now the `ati` has changed to the new `AccessToken` ID.

The original `RefreshToken` and the new one returned on the `AccessToken` refresh is different based on the `ati`.

I'm very reluctant on changing the structure of the JWT to not include the `AccessToken` id reference but don't see any other option to resolve this issue at this time.

Just to add...this issue does not allow JWT `RefreshToken`'s to be reused.

Any thoughts on your end?

dsyer:
It's a little bit perverse to use JWT and not the `JwtTokenStore`. Can someone explain why you need an external store if the tokens have all their data encoded in them? (If you don't use an external store the ati is just a memo for the client, and has no significance whatsoever in the server.)

maxim-filkov:
> As a quick workaround, each time you refresh an AccessToken, ensure you use the returned RefreshToken on the next refresh_token grant flow. You will also need to ensure that reuseRefreshToken = false for DefaultTokenServices to make this work.

> Please give this workaround a try and let me know if it works on your end. I have a test on my end that works.

Hi Joe, this should definitely work. The only small issue I have, I can't set reuseRefreshToken to false, this is how I do this:

```
@Configuration
@EnableAuthorizationServer
protected static class AuthorizationServerConfiguration extends AuthorizationServerConfigurerAdapter {
    ...
    @Bean
    @Primary
    @Transactional
    public AuthorizationServerTokenServices tokenServices() {
       final DefaultTokenServices tokenServices = new DefaultTokenServices();
       tokenServices.setSupportRefreshToken(true);
       tokenServices.setReuseRefreshToken(false); // here
       tokenServices.setTokenStore(tokenStore());
       return tokenServices;
    }
```
While debugging in the DefaultTokenServices (line 171) the value is still *true* here:
```
if (!reuseRefreshToken) {
    tokenStore.removeRefreshToken(refreshToken);
    refreshToken = createRefreshToken(authentication);
}
```
It's a different instance of DefaultTokenServices also that is used during runtime. Not the one I'm initialising in the configuration class. **Could you please hint here what I'm doing wrong?**
But in any way I created triggers in database to keep the token tables in consistent state and the workaround does work.

maxim-filkov:
> It's a little bit perverse to use JWT and not the JwtTokenStore. Can someone explain why you need an external store if the tokens have all their data encoded in them? (If you don't use an external store the ati is just a memo for the client, and has no significance whatsoever in the server.)

Hi Dave, I need external story because in this case I can revoke refresh tokens if needed. And if I use external token store without JWT, the tokens look like "f50151cb-f247-4dda-a05c-d357878b22ac" so they lack the payload.

dsyer:
@maxim-filkov I see. Isn't it more intuitive for users to manage approvals (which represent a concrete decision they made) as opposed to refresh tokens (kind of abstract). That was the way it was intended to work anyway, so you don't need a token store, but you might want an approval store when you switch to JWT tokens.

maxim-filkov:
Dave, thank you for the clarification. I'll switch back to JwtTokenStore then to keep the things simpler.

jgrandja:
Hi @maxim-filkov, can you confirm if I can close this PR given you are switching to `JwtTokenStore`?

maxim-filkov:
Hi Joe, yes, thanks to Dave it's clear now I had a corner case. I'm switching back to JwtTokenStore. Thank you!

jgrandja:
Thanks for letting me know @maxim-filkov.

Issue #810, opened 2016-07-22
Authorization Server and OAuth2 Client don't agree on HTTP status codes


shanman190:
I've setup a basic test showcasing the issue that I'm seeing and the results that I would have expected. The test case that fails specifically is the `access_admin_forbidden` test. The gist is that the server returns a 403 response code from the denied access, but when the client is processing the response it creates a UserDeniedAuthorizationException whose status code is set to 400. There appears to be a discrepancy, but I'd like to solicit some feedback before creating a PR.

```java
package demo.oauth;

import org.junit.Test;
import org.junit.runner.RunWith;
import org.springframework.beans.factory.annotation.Value;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
import org.springframework.boot.test.SpringApplicationConfiguration;
import org.springframework.boot.test.TestRestTemplate;
import org.springframework.boot.test.WebIntegrationTest;
import org.springframework.context.annotation.Configuration;
import org.springframework.http.HttpStatus;
import org.springframework.http.ResponseEntity;
import org.springframework.security.config.annotation.web.builders.HttpSecurity;
import org.springframework.security.oauth2.client.OAuth2RestTemplate;
import org.springframework.security.oauth2.client.token.grant.client.ClientCredentialsResourceDetails;
import org.springframework.security.oauth2.config.annotation.configurers.ClientDetailsServiceConfigurer;
import org.springframework.security.oauth2.config.annotation.web.configuration.AuthorizationServerConfigurerAdapter;
import org.springframework.security.oauth2.config.annotation.web.configuration.EnableAuthorizationServer;
import org.springframework.security.oauth2.config.annotation.web.configuration.EnableResourceServer;
import org.springframework.security.oauth2.config.annotation.web.configuration.ResourceServerConfigurerAdapter;
import org.springframework.security.oauth2.config.annotation.web.configurers.AuthorizationServerSecurityConfigurer;
import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
import org.springframework.web.bind.annotation.RequestMapping;
import org.springframework.web.bind.annotation.RestController;

import static org.junit.Assert.assertEquals;

@RunWith(SpringJUnit4ClassRunner.class)
@WebIntegrationTest(randomPort = true)
@SpringApplicationConfiguration(classes = DemoOAuth2Application.class)
public class DemoOAuth2ApplicationTests {
    @Value("${local.server.port}")
[...]

PR #868, opened 2016-10-06
Correct OAuth2 deserializers to handle OAuth server returning access denied errors

shanman190:
Update OAuth2 exception deserializers to deserialize access denied exceptions into the OAuth2AccessDeniedException. By merging this change the OAuth server and OAuth client will agree on HTTP status code 403 when the OAuth client is denied access to perform the requested action.

Fixes gh-810


jgrandja:
Closing this PR as it's been fixed in #875.

PR #871, opened 2016-10-07
Always sort scopes before creating the MD5 checksum

### LINKED ISSUE #874: Inconsistent generation of `authentication_id`
If you subclass `DefaultAuthenticationKeyGenerator` and override the method `extractKey(...)` and make it return a debuggable key like below: 
```Java
Map<String, String> values = new LinkedHashMap<String, String>();
OAuth2Request authorizationRequest = authentication.getOAuth2Request();
if (!authentication.isClientOnly()) {
   values.put(USERNAME, authentication.getName());
}
values.put(CLIENT_ID, authorizationRequest.getClientId());
if (authorizationRequest.getScope() != null) {
  values.put(SCOPE, OAuth2Utils.formatParameterList(authorizationRequest.getScope()));
}
return values.toString();
```

You will see the `scopes` are some times out of order so the `authentication_id` will have a different checksum value.

The is caused by the `BaseRequest.getScope()` always returning a `HashSet`. From my perspective the proper fix is to always sort the scopes when generating the id so regardless what kind of collection we are given the results would remain the same.

born2snipe:
This fixes #874 where authorization ids could be generated differently if there was more then one scope.  This was due to the scopes being stored in a set but not sorted prior to hashing the values.  

For example:
```java
// Given these sets
Set<String> a = new LinkedHashSet<String>(Arrays.asList("a", "c", "b"));
Set<String> b = new LinkedHashSet<String>(Arrays.asList("c", "a", "b"));

// this would be true
Assert.assertEquals(a, b);

// due to order differences 
Assert.assertNotEquals(OAuth2Utils.formatParameterList(a), OAuth2Utils.formatParameterList(b)); 

// now being sorted the string will be equal
Assert.assertEquals(OAuth2Utils.formatParameterList(new TreeSet(a)), OAuth2Utils.formatParameterList(new TreeSet(b))); 

```

If you subclass `DefaultAuthenticationKeyGenerator` and override the method `extractKey(...)` and make it return a debuggable key like below: 
```Java
Map<String, String> values = new LinkedHashMap<String, String>();
OAuth2Request authorizationRequest = authentication.getOAuth2Request();
if (!authentication.isClientOnly()) {
   values.put(USERNAME, authentication.getName());
}
values.put(CLIENT_ID, authorizationRequest.getClientId());
if (authorizationRequest.getScope() != null) {
  values.put(SCOPE, OAuth2Utils.formatParameterList(authorizationRequest.getScope()));
}
return values.toString();
```

You will see the `scopes` are some times out of order so the `authentication_id` will have a different checksum value.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/MapDigester.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/MapDigester.java
@@ -1,0 +1,27 @@
+package org.springframework.security.oauth2.common.util;
+
+import java.io.UnsupportedEncodingException;
+import java.math.BigInteger;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Map;
+
+public class MapDigester {
+    public String digest(Map values) {
+        MessageDigest digest;
+        try {
+            digest = MessageDigest.getInstance("MD5");
+        }
+        catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("MD5 algorithm not available.  Fatal (should be in the JDK).");
+        }
+
+        try {
+            byte[] bytes = digest.digest(values.toString().getBytes("UTF-8"));
+            return String.format("%032x", new BigInteger(1, bytes));
+        }
+        catch (UnsupportedEncodingException e) {
+            throw new IllegalStateException("UTF-8 encoding not available.  Fatal (should be in the JDK).");
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


benkiefer:
Any update on this? We are seeing unnecessary creation of additional access and refresh tokens because of this issue and duplicate key exceptions as well.

jgrandja:
Hi @benkiefer I'll be getting to this shortly.

jgrandja:
Thanks for the PR @born2snipe!

FYI, I added some polish as I felt it wasn't necessary to have a separate class for the digest calculation so I brought it back into `DefaultAuthenticationKeyGenerator`.

Issue #807, opened 2016-07-12
JwtTokenStore doesn't work properly in combination with custom DB approval store


GFriedrich:
I tried to use a custom approval store to store them inside a DB.
Problem is: Whenever an access token is expired or a refresh token is used, the stored approvals are removed by the JwtTokenStore (and never added back). So as soon as the next access/refresh token is used, the token is rejected, as the DB approval store can't find any appovals.
I'm not sure why it's implemented in this way, as this removes the purpose (at least for me) to use a custom approval store.
So either this is a bug or I'm using the approval store in a wrong way.

dsyer:
Coincidentally, I was scratching my head about this one today as well. I'm happy to believe it's a bug in `JwtTokenStore`, but that code was copied over from Cloud Foundry, where there don't seem to be any issues. Maybe there was a transcription error (since the code is not identical). I can see why approvals should be removed when a refresh token expires, but not when it is redeemed. My best guess is that the intention is to remove the approvals, and then add them back, in case the refresh grant has narrowed the scopes.

wfaithfull:
I'm having this issue, so I'll just think aloud in here.

So, my understanding of all the plumbing involved is still a little shaky, but I'll do my best. I believe the approvals are added in 

`ApprovalStoreUserApprovalHandler#updateAfterApproval` 

(assuming that one is configured). The method in question is triggered after a request to `/oauth/authorize`.

The issue is that, when an access token expires, `DefaultTokenServices#loadAuthentication`  calls `tokenStore.removeAccessToken(accessToken)`, which in `JwtTokenStore`, removes the approvals. When `DefaultTokenServices#refreshAccessToken` calls `JwtTokenStore#readRefreshToken`, the currently approved scopes are compared to the requested scopes. Since there are no approved scopes now, it returns null.

GFriedrich:
FYI: I stopped thinking about this problem one day after submitting this issue as I found a workaround: I added a wrapper object around my custom approval store, which only calls the "get" methods of the approval store. The other methods are a no-op. This wrapper object is handed over to the JwtTokenStore, so the token store is no longer able to modify any of my stored tokens. It's a bit hacky, but works as expected :)

wfaithfull:
@GFriedrich doesn't that mean that your approvals last forever? Or do you have some other means by which you remove them upon refresh token expiry?

GFriedrich:
[...]

PR #877, opened 2016-10-12
Fix autoapprovals not being added to approval store

wfaithfull:
#807 was marked as fixed, but there is still an issue if the scopes are autoapproved - the approvals are never actually inserted into the `ApprovalStore` if _all_ scopes are auto approved, which causes the refresh token to result in an `InvalidGrantException` because there are no approvals in the store.


jgrandja:
Thank you for the PR @wfaithfull !
This has now been merged to master.

wfaithfull:
Thank you!

PR #903, opened 2016-11-16
SECOAUTH-142 - Fix storing access tokens in HTTP session

dpeykov:
The current implementation of `HttpSessionOAuthRememberMeServices` does not allow access tokens to be stored, despite its claim that it "stores everything" and that "storing access tokens can be suppressed".

The condition `storeAccessTokens && !token.getValue().isAccessToken()` will never be `true` for an access token, and setting the `storeAccessTokens` boolean to `false` would prevent all tokens from being saved and break the functionality entirely. The original intent was probably `storeAccessTokens || !token.getValue().isAccessToken()`.

This PR fixes this by selectively storing access tokens based on the `storeAccessTokens` boolean, and exposing a getter/setter for it. Note that the default value is switched to `false` to match the current behavior which only stores request tokens.

### DIFF
--- a/spring-security-oauth/src/main/java/org/springframework/security/oauth/consumer/rememberme/HttpSessionOAuthRememberMeServices.java
+++ b/spring-security-oauth/src/main/java/org/springframework/security/oauth/consumer/rememberme/HttpSessionOAuthRememberMeServices.java
@@ -10,9 +10,10 @@ import javax.servlet.http.HttpSession;
 import org.springframework.security.oauth.consumer.OAuthConsumerToken;
 
 /**
- * Default implementation of the OAuth2 rememberme services. Just stores everything in the session by default. Storing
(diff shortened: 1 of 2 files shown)


jgrandja:
@dpeykov Thanks for the PR.

> The current implementation of HttpSessionOAuthRememberMeServices does not allow access tokens to be stored

This does not seem correct. If you look at the `HttpSessionOAuthRememberMeServicesTests.testStoreEverything()`, the token(s) are being saved in the `HttpSession`.

I'm going to close this PR as I don't see any issue here. If you believe there is an issue, please log a new ticket describing the bug along with a test that demonstrates the bug.

dpeykov:
`testStoreEverything()` is broken: it creates two tokens and then asserts that only one was stored (i.e. not everything).

It is a duplicate of `testStoreRequestTokensOnly()` because storing access tokens doesn't work (hence this PR).

PR #919, opened 2016-12-06
New TestRule and @WithSecurityContext annotation for mocking OAuth2 s…

claassen:
…copes during tests.


claassen:
New test helpers similar to the ones here: https://github.com/spring-projects/spring-security/tree/master/test/src/main/java/org/springframework/security/test/context/support but for mocking OAuth2 scopes.

PR #928, opened 2016-12-16
fix-JWT-refresh-tokens-unable-to-refresh-themselves

gharizanov92:
This change is needed because JwtTokenStore has no implementation for storeAccessToken and storeRefreshToken.
Which is fine, because JWT doesn't require persistence. But a refresh token is valid if
1) Digital signature is verified
2) There's an approval in the DB for all the scopes and the client_id, stored in the JWT itself.

When you refresh a token, the DefaultTokenServices removes all refresh tokens and access tokens. The only reasonable
way to 'remove' a JWT is to remove it's approval in the approval store. Since once removed it's never inserted again (user has to grant an approval again)
the next time you use a refresh token it will not be validated because of 2) and will fail.
A simple fix for that is providing an implementation for store(Access/Refresh)Token which just inserts an approval to the DB.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/JwtTokenStore.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/JwtTokenStore.java
@@ -12,18 +12,19 @@
  */
 
 package org.springframework.security.oauth2.provider.token.store;
-
 import org.springframework.security.core.Authentication;
 import org.springframework.security.oauth2.common.DefaultExpiringOAuth2RefreshToken;
 import org.springframework.security.oauth2.common.DefaultOAuth2RefreshToken;
 import org.springframework.security.oauth2.common.OAuth2AccessToken;
 import org.springframework.security.oauth2.common.OAuth2RefreshToken;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@gharizanov92 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=928)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=928) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@gharizanov92 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=928)!

jgrandja:
Thanks for the PR @gharizanov92. There are a few failing tests that need to be resolved.

If you still feel this issues needs to be addressed, please fix the failing tests and add test(s) for the proposed changes.

jgrandja:
@gharizanov92 I'm closing this PR due to lack of feedback from review.

If you still feel this is an issue, please submit a new PR and address the review feedback.

Issue #935, opened 2016-12-21
Redis based AuthorizationCodeServices


srempfer:
In a project on that I currently working we use the Redis based TokenStore.
To be able to completely scale out the Authorization Server with Redis it would be great to have a Redis based AuthorizationCodeServices as well.

matejsp:
Is there any update regarding Redis support for AuthorizationCodeServices?

srempfer:
I think currently the manpower goes into Spring Security 5.0 changes to consolidate the OAuth stuff. See https://github.com/spring-projects/spring-security/issues/3907

subaan:
Is there any update regarding Redis support?

zenglongx:
Is there any update regarding Redis support for AuthorizationCodeServices?

Jeff-Walker:
Anyone have any ideas on the best way to implement one?

PR #936, opened 2016-12-21
Add redis based authorization code services and tests

srempfer:
Fixes gh-935

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/code/RedisAuthorizationCodeServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/code/RedisAuthorizationCodeServices.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2006-2016 the original author or authors.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


nasvillanueva:
Hope this gets merged soon. For the mean time, I'll be including the fix to our code base.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/code/RedisAuthorizationCodeServices.java
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	@Override
+	protected void store(String code, OAuth2Authentication authentication) {
+		byte[] key = serializeKey(AUTH_CODE + code);
+		byte[] auth = serialize(authentication);
+
+		RedisConnection conn = getConnection();
+		try {
+			conn.set(key, auth);
```
sankin: By default the code is a randomly generated 6-letter string. It's possible to generate two identical codes, especially in a system that handles a high degree of volume. 

This implementation of the `store` method will quietly overwrite the existing code stored in Redis. I believe a better approach is to use SETNX command instead (https://redis.io/commands/setnx) and verify that the code is actually persisted without collision (you get a `Boolean` from setNX call see https://docs.spring.io/spring-data/redis/docs/current/api/org/springframework/data/redis/connection/RedisStringCommands.html#setNX-byte:A-byte:A-).

jgrandja:
Thanks for the PR @srempfer. This is now in master.

Issue #949, opened 2017-01-16
unable to use RSA public key without comment (id)


mzeibig:
In `org.springframework.security.jwt.crypto.sign.RsaKeyHelper#parsePublicKey` a key without comment only matches if it has a trailing space.
This methot is called in `org.springframework.security.jwt.crypto.sign.RsaVerifier#RsaVerifier(java.lang.String)`, but a trailing space is removed by calling `trim()` on the key string.

PR #950, opened 2017-01-16
fix for #949

mzeibig:
Added unit test and changed regexp for #949


jgrandja:
Thanks for the PR @mzeibig. This is now in master.

Issue #974, opened 2017-02-09
Allow optional parameters for introspection request


mouellet:
In [RFC 7662](https://tools.ietf.org/html/rfc7662#section-2.1) :

> The introspection endpoint MAY accept other OPTIONAL parameters to
   provide further context to the query...

To support this, the `RemoteTokenService` needs to add those optional parameters to the introspection request. As proposed in the standard, one optional parameter is to be called `token_type_hint`. Also, the request should sent parameters as "application/x-www-form-urlencoded" data, so adding them to the existing `formData` in `loadAuthentication()` seems like a simple solution.

EtienneK:
I second this.

Also, RemoteTokenServices must check for the "active" key (with true or false value) in the response and allow/fail accordingly: according to RFC7662 (https://tools.ietf.org/html/rfc7662), the introspection endpoint will always return HTTP status 200 OK, even if the token isn't valid. Thus, simply checking for a 400 response status is not enough.

PR #990, opened 2017-02-25
Add optional parameters in RemoteTokenService

mouellet:
Support for introspection request (see RFC 7662).

Fixes gh-974


EtienneK:
When will this be implemented? It is needed to comply with RFC 7662 (https://tools.ietf.org/html/rfc7662)

jgrandja:
@mouellet If you are still looking to add this enhancement, can you please add a test and we'll look at getting this merged.

jgrandja:
Thanks for the PR @mouellet. This is now in master.

PR #1011, opened 2017-03-08
Fix #1010 'RSA Signature did not match content' with valid RSA key

### LINKED ISSUE #1010: 'RSA Signature did not match content' while validating token signature with valid RSA key
## Symptom
Given a JSON Web Token
And using _security.oauth2.resource.jwk.key-set-uri_ property to load public RSA key
When validating the token signature with the public key
Then the validation failed with the message 'RSA Signature did not match content'

## Diagnosis
In method org.springframework.security.oauth2.provider.token.store.jwk.JwkDefinitionSource#createRsaVerifier RSA key is created starting from **modulus** and **exponent** as signed integer instead of _unsigned_ int like specified in RFC 7518 (6.3.1)

gwallet:
Fix #1010 by extracting **modulus** and **exponent** as _unsigned_ int


jgrandja:
Thanks for the PR @gwallet. I applied some minor polish and merged to master.

PR #1020, opened 2017-03-16
RemoteTokenServices should not check for client_id

gonzalad:
PR for issue #838 

Per RFC 6749 (which doesn't address the topic)
and RFC 7662 (which explicitly permits this),
a valid check token response from the authorization
server might not contain a client_id.

This fix adress that, using the client_id used
for calling the introspect endpoint to avoid
impacting DefaultAccessTokenConverter.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
@@ -111,7 +111,9 @@ public class RemoteTokenServices implements ResourceServerTokenServices {
 			throw new InvalidTokenException(accessToken);
 		}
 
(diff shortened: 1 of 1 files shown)


jgrandja:
Hi @gonzalad I just reviewed your PR and the assignment of `RemoteTokenServices.clientId` to the `Map` returned from the _Introspection Response_ is not correct. `RemoteTokenServices.clientId` is the client credentials used to authenticate with the _Authorization Server_ in order to access the _Introspection Endpoint_. The `client_id` that may optionally be in the `Map` returned from the _Introspection Response_ is the `client_id` that was originally authorized by the _Resource Owner_.
I'm going to close this PR but will be addressing this issue given the fact that `client_id` is an optional parameter in the _Introspection Response_.

gonzalad:
Hi @jgrandja, yes I completely missed that point
Sorry !

Issue #958, opened 2017-01-22
RemoteTokenServices throws IllegalStateException instead of InvalidTokenException (rfc7662)


gonzalad:
Hello,

I'm using RemoteTokenServices in my OAuth Resource Server to validate OAuth 2.0 bearer tokens by callling the introspect endpoint of my Authorization Server (https://tools.ietf.org/html/rfc7662).

From the spec, https://tools.ietf.org/html/rfc7662#section-2.3 (see [1]), when introspecting an expired or invalid access token, the server responds with 
`
{
  "active": false
}
`

Problem is that Spring relies on `error` attribute and not `active` attribute.

https://github.com/spring-projects/spring-security-oauth/blob/ec215f79f4f73f8bb5d4b8a3ff9abe15b3335866/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java#L109

Could this code be updated to handle this case ?

Thanks,
Adrian

[1] Quote from rfc7662#section-2.3:
> Note that a properly formed and authorized query for an inactive or
>    otherwise invalid token (or a token the protected resource is not
>    allowed to know about) is not considered an error response by this
>    specification.  In these cases, the authorization server MUST instead
>    respond with an introspection response with the "active" field set to
>    "false" as described in Section 2.2.

adnanalawiyat:
This is also affecting my project. Please fix

adnanalawiyat:
would it be acceptable to fix by adding check for active = false in org.springframework.security.oauth2.provider.token.RemoteTokenServices#loadAuthentication. Ex:

```java
if (map.containsKey("error")) {
	logger.debug("check_token returned error: " + map.get("error"));
	throw new InvalidTokenException(accessToken);
}
```

becomes: 

```java
if (map.containsKey("error")
			|| (map.containsKey("active") && "false".equals(map.get("active")))) {
	logger.debug("check_token returned error: " + map.get("error"));
	throw new InvalidTokenException(accessToken);
}
```

gonzalad:
Hello,
PR available #1021 
Could it be considered for inclusion please ?
Thanks,
Adrian

jgrandja:
Resolved via f1a9b97.
See [comment](https://github.com/spring-projects/spring-security-oauth/issues/838#issuecomment-301554015) in #838

PR #1021, opened 2017-03-16
958: RemoteTokenServices should handle active

gonzalad:
PR for issue #958 

RTS should handle active attribute in response
as per rfc7662.

active=false means the token is expired and now
raises a InvalidTokenException.


jgrandja:
Resolved via f1a9b97.
See [comment](https://github.com/spring-projects/spring-security-oauth/issues/838#issuecomment-301554015) in #838

PR #1040, opened 2017-04-15
Add isDebugEnabled

kazuki43zoo:
In current implementation, there are case that generate an unnecessary string instance at some debug logs.


jgrandja:
Thanks for the PR @kazuki43zoo. This is now in master.

Issue #1058, opened 2017-05-04
An issue with grant_type password and a trusted client


luchob:
I would like to get an access token with a password grant as described [here](https://aaronparecki.com/oauth-2-simplified/#others), for example. This means that a similar request will be issued to get a token:

```
POST https://myserver.com/oauth/token
      grant_type=password&
      username=USERNAME&
      password=PASSWORD&
      client_id=CLIENT_ID
```

This seems OK and is hinted in the [RFC](https://tools.ietf.org/html/rfc6749#page-39). To implement all this it seems that I just should call `allowFormAuthenticationForClients()` of `AuthorizationServerSecurityConfigurer` which essentially enables `ClientCredentialsTokenEndpointFilter` in the filter chain. (a small javadoc would be very nice :-) ) See a sample configuration [here](https://github.com/luchob/oauth2-passwordgrant-example/blob/demo1/auth/src/main/java/eu/balev/demo/auth/config/OAuth2Config.java#L63-L69). Now the problems:

**Problem 1.** I'm forced to permit the check access token end point ( `checkTokenAccess("permitAll()")` ) for anyone because the `/oauth/check_token` endpoint is not considered into `ClientCredentialsTokenEndpointFilter`. That is because of the internal class `ClientCredentialsRequestMatcher`. IMHO this is a bug - why should the filter be available only for the `/oauth/token` endpoint and what am I supposed to do with a token that can't be verified?

**Problem 2.** This comes from the previous problem. If I open the check token endpoint (which was not recommended by @dsyer at StackOverflow) then I'm running into the next issue, namely [this one](https://github.com/spring-projects/spring-security-oauth/issues/717).

Is this really a bug or I miss something?

dsyer:
Fair enough I guess. Some comments:

* You don't need to `allowFormAuthenticationForClients()` (and the RFC recommends that you don't). It's better to use HTTP Basic for client authentication.

* Form-based authentication would be equally bad for /check_token, but if you want to enable it you could add your own filter. Enabling it with `allowFormAuthenticationForClients()` would be a new feature, but it probably won't be a high priority, unless there is a PR.

* There's nothing forcing you to use `RemoteTokenServices`, and nothing preventing you from customizing it (as I pointed out in #717).

luchob:
Dave, thanks for the quick answer!

[...]

PR #1061, opened 2017-05-05
 An issue with grant_type password and a trusted client

luchob:
Based on our discussion in #1058 

That is how imagine the feature. I think that if you enable willingly the form auth it should apply to all endpoints.
I hope I've kept the contribution guidelines, although I used something from 1.7 initially :-)

Thank you!

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/annotation/web/configurers/AuthorizationServerSecurityConfigurer.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/annotation/web/configurers/AuthorizationServerSecurityConfigurer.java
@@ -16,6 +16,7 @@
 package org.springframework.security.oauth2.config.annotation.web.configurers;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/annotation/web/configurers/AuthorizationServerSecurityConfigurer.java
```diff
@@ -205,8 +214,14 @@ public void configure(HttpSecurity http) throws Exception {
 	}
 
 	private ClientCredentialsTokenEndpointFilter clientCredentialsTokenEndpointFilter(HttpSecurity http) {
+		
+		FrameworkEndpointHandlerMapping handlerMapping = frameworkEndpointHandlerMapping();
+
 		ClientCredentialsTokenEndpointFilter clientCredentialsTokenEndpointFilter = new ClientCredentialsTokenEndpointFilter(
-				frameworkEndpointHandlerMapping().getServletPath("/oauth/token"));
+				Arrays.asList(handlerMapping.getServletPath("/oauth/token"),
+						handlerMapping.getServletPath("/oauth/token_key"),
```
jgrandja, line 222: This is not correct. The `ClientCredentialsTokenEndpointFilter` should only match on `/oauth/token`. The default authorization server configuration will setup the authorization rules for `/oauth/token_key` and `/oauth/check_token`.

jgrandja:
Closing due to lack of feedback requested.

PR #1073, opened 2017-05-22
Fix to allow double exp values

tomcruise81:
A `NumericDate` is described in the [Terminology](https://tools.ietf.org/html/rfc7519#section-2) section of [RFC7519](https://tools.ietf.org/html/rfc7519). Based on that description, it is not mandatory that these values be integral (i.e. they can contain decimal places).

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
@@ -104,7 +104,13 @@ public class DefaultAccessTokenConverter implements AccessTokenConverter {
 		info.remove(CLIENT_ID);
 		info.remove(SCOPE);
 		if (map.containsKey(EXP)) {
-			token.setExpiration(new Date((Long) map.get(EXP) * 1000L));
+			Object expObject = map.get(EXP);
+			if (expObject instanceof Double) {
+				token.setExpiration(new Date((long) ((Double) expObject * 1000D)));
+			}
+			else if (expObject instanceof Long) {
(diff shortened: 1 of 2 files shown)


tomcruise81:
Any possibility of getting this one merged in?

jgrandja:
@tomcruise81 RFC7519 defines `NumericDate` as follows:

> NumericDate
>       A JSON numeric value representing the number of seconds from
>       1970-01-01T00:00:00Z UTC until the specified UTC date/time,
>       ignoring leap seconds.  This is equivalent to the IEEE Std 1003.1,
>       2013 Edition [POSIX.1] definition "Seconds Since the Epoch", in
>       which each day is accounted for by exactly 86400 seconds, other
>       than that non-integer values can be represented.  See RFC 3339
>       [RFC3339] for details regarding date/times in general and UTC in
>       particular.

It specifically states **number of seconds** since epoch. I do not see where it states that it may contain decimal places?

Is this causing an issue for one of your use cases?

tomcruise81:
> other than that non-integer values can be represented.

Our security gateway/token provider's implementation took that to mean that they can incorporate milliseconds using decimal places.

tomcruise81:
If the general consensus is that the appropriate means of handling this is for them to truncate their milliseconds, I can re-approach the provider...

jgrandja:
@tomcruise81 The spec does not explicitly state that decimals may be contained so I'm not inclined to merge this as I haven't seen this issue reported by any other user. The most common format for a time is expressed in milliseconds as a `long` not `double`.
I would approach the provider to resolve this for you.

jgrandja:
I'm going to close this but if you still have this issue after approaching the provider than perhaps there is an alternative solution we can come up with.

PR #1087, opened 2017-06-15
Handle the presence of x5c x5t in JWK

### LINKED ISSUE #1082: JwkSetConverter should handle arrays e.g x5c
We're trying to use Auth0, but JwkSetConverter is only set up to handle strings, yet x5c is an array (as specified at https://tools.ietf.org/html/rfc7515#section-4.1.6), so the parsing blows up.

Is the workaround as simple as fix as it may appear (skip the x5 fields)?

nealeu:
This is sufficient to resolve #1082

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkAttributes.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkAttributes.java
@@ -55,6 +55,19 @@ final class JwkAttributes {
 	static final String RSA_PUBLIC_KEY_EXPONENT = "e";
 
 	/**
+	 * The &quot;x5c&quot; (X509 certificate chain) parameter contains the X.509 public key certificate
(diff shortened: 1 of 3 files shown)


jgrandja:
Thanks for the PR @nealeu. I added some polish and merged to master.

nealeu:
Thanks for doing so promptly after the obvious Spring 5-related work.  It's good timing as I got my Auth0 with JWK stuff working today with a bit too much copy-paste of closed classes.  I'm hoping you'll be able to make life a bit easier for us on that front, as it's really important for us with the automatic key rotation.

jgrandja:
@nealeu As an FYI, we're not planning on opening up the JWK-related classes as we're not looking to build too much more support than what is already provided. Our intention is to bring the JWK feature to Spring Security 5 and build full support there instead. As of now, our focus is on the client side and next will be the Resource Server side which will entail the JWK feature.

nealeu:
Thanks for the FYI.  We can get by with a bit of copy paste for what's closed.  Good to hear focus on future release with all the reactive goodies it brings :)

PR #1095, opened 2017-06-22
When using Jackson to serialize OAuth2AccessToken, it will never expire, fix this problem.

xyrc:


jgrandja:
Hi @xyrc. As per spec, a [Successful Access Token Response](https://tools.ietf.org/html/rfc6749#section-5.1) contains the `expires_in` attribute:

>    expires_in
>          RECOMMENDED.  The lifetime in seconds of the access token.  For
>          example, the value "3600" denotes that the access token will
>          expire in one hour from the time the response was generated.
>          If omitted, the authorization server SHOULD provide the
>          expiration time via other means or document the default value.

It's the responsibility of the client to keep track of when the Access Token expires. For example, when the client receives the Access Token response, it will know when the Access Token expires based on this calculation:

expiration time = issued time  + (expires_in / 60) in mins

** issued time is when the client receives the access token

Also, based on your PR, you added the new attribute `expiration_time`, which is not a valid attribute for an Access Token response, as defined by the spec.

I hope this clarifies things?

xyrc:
The problem is that the serialization results cannot be stored.
 In the test case, we mock the System.currentTimeMillis () to a fixed value, this is not true; 
the true is: 
you can instantiate a DefaultOAuth2AccessToken object, Store it in JSON format, after few minutes, Deserialization it, Will get different object, 
because:
expiration_time = currentTime + (expires_in * 1000.)

jgrandja:
Hi @xyrc. I see your point regarding deserializing `DefaultOAuth2AccessToken` on the client side, given the way `setExpiresIn` is implemented with the `System.currentTimeMillis()`.

However, even though the `expiration` may not be correct at deserialization time, if the access token is in fact expired, it will get rejected by Resource Servers when they validate the incoming access token via either a callback to the Authorization server or if it's a JWT with the `exp` claim.

Maybe you can propose another solution that doesn't involve adding the `expiration_time` attribute to the Access Token response?...as this is not part of the spec

PR #1104, opened 2017-06-27
JwkTokenStore Supports Multiple JWK Set URLs

rwinch:


spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkTokenStore.java
```diff
@@ -86,16 +88,23 @@
  * @author Joe Grandja
  */
 public final class JwkTokenStore implements TokenStore {
-	private final JwtTokenStore delegate;
```
rwinch: Is that true? Couldn't another token store do the exact same thing? Remember we don't allow consumers to set this value anyways, so this is of no risk but keeps the remaining code as loosely coupled as possible.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkDefinitionSource.java
```diff
@@ -90,7 +100,9 @@ JwkDefinition getDefinitionLoadIfNecessary(String keyId) {
 			return result;
 		}
 		this.jwkDefinitions.clear();
-		this.jwkDefinitions.putAll(loadJwkDefinitions(this.jwkSetUrl));
+		for(URL jwkSetUrl : jwkSetUrls) {
```
rwinch: @jgrandja We already have a race condition between the clear and putAll and was saving that for a separate issue. I pushed 2 more commits to address #1105

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkTokenStore.java
jgrandja: Ok agreed there could be another Jwt specific implementation of `TokenStore`. And although the member variable is completely hidden in the class, it's always best practice to reference against the interface.

PR #1114, opened 2017-07-11
Implement Issuer claim verification

jgrandja:


spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/discovery/ProviderDiscoveryClient.java
```diff
@@ -0,0 +1,97 @@
...
+			builder.userInfoEndpoint(this.toURL((String)responseAttributes.get(USERINFO_ENDPOINT_ATTR_NAME)));
+		}
+		if (responseAttributes.containsKey(JWK_SET_URI_ATTR_NAME)) {
+			builder.jwkSetUri(this.toURL((String)responseAttributes.get(JWK_SET_URI_ATTR_NAME)));
+		}
+
+		return builder.build();
+	}
+
+	private URL toURL(String urlStr) {
```
rwinch: Perhaps the Builder should handle a String and then this can be removed.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/JwtAccessTokenConverter.java
```diff
@@ -75,6 +65,8 @@
 
 	private AccessTokenConverter tokenConverter = new DefaultAccessTokenConverter();
 
+	private JwtClaimsSetVerifier jwtClaimsSetVerifier;
```
rwinch: Rather than doing null checks, create a `NoOpJwtClaimsSetVerifer` (private scope) and does nothing for verification

rwinch:
I would also create a `DelegatingJwtClaimsSetVerifier` (perhaps in another PR if you like but can be in this one) which has a `List<JwtClaimsSetVerifier>` and invokes each of them. This ensures that multiple `JwtClaimsSetVerifier` can be easily used for users with custom claim verification

jgrandja:
Thanks for the review @rwinch. All feedback has been applied.

PR #1128, opened 2017-08-01
Fix issue for some OAuth2 client who are still sending refreshToken 

stevehouel:
Fix issue for some OAuth2 client who are still sending refreshToken even if grant type is settled to ClientCredentials (ex Keycloak). This cause an issue when access token expired.

Regarding to RFC original code was correct but to prevent further issues this PR fix and force bypass usage of refresh token when using client credentials mode.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/token/AccessTokenProviderChain.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/token/AccessTokenProviderChain.java
@@ -35,7 +35,7 @@ import org.springframework.security.oauth2.common.exceptions.OAuth2Exception;
  * A chain of OAuth2 access token providers. This implementation will iterate through its
  * chain to find the first provider that supports the resource and use it to obtain the
  * access token. Note that the order of the chain is relevant.
- * 
+ *
  * @author Ryan Heaton
  * @author Dave Syer
  */
@@ -53,7 +53,7 @@ public class AccessTokenProviderChain extends OAuth2AccessTokenSupport
 
 	/**
 	 * Token services for long-term persistence of access tokens.
-	 * 
+	 *
 	 * @param clientTokenServices the clientTokenServices to set
 	 */
 	public void setClientTokenServices(ClientTokenServices clientTokenServices) {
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@stevehouel Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1128)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1128) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@stevehouel Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1128)!

jgrandja:
@stevehouel Thanks for the PR. Can you please provide a test for this case.

stevehouel:
Sure I do it asap.
Le lun. 14 août 2017 à 23:08, Joe Grandja <notifications@github.com> a
écrit :

jgrandja:
Thanks for the PR @stevehouel. This is now in master.

stevehouel:
Thx @jgrandja

hbergmey:
In my opinion this change is partly invalid.

We are currently facing an issue, where this proposed fix, especially the added "!isClientOnly" check in line 108 of AccessTokenProviderChain prevents the use of refresh tokens in the client scenario without user sessions all together.

` // lines 96 to 116 in optainAccesToken()
		if (resource.isClientOnly() || (auth != null && auth.isAuthenticated())) {  // p1
			existingToken = request.getExistingToken();
			if (existingToken == null && clientTokenServices != null) {
				existingToken = clientTokenServices.getAccessToken(resource, auth);
			}

			if (existingToken != null) {
				if (existingToken.isExpired()) {
					if (clientTokenServices != null) {
						clientTokenServices.removeAccessToken(resource, auth);
					}
					OAuth2RefreshToken refreshToken = existingToken.getRefreshToken();
					if (refreshToken != null && !resource.isClientOnly()) {  // p2
						accessToken = refreshAccessToken(resource, refreshToken, request);
					}
				}
				else {
					accessToken = existingToken;
				}
			}
		}
`

As you can see, at p1 'resource.isClientOnly()' must return true, because auth will not be defined without user sessions. But at p2  '!resource.isClientOnly()' always prevents the execution of refreshAccessToken. Thus the access token can never be refreshed.

The addition of '&& !resource.isClientOnly()' seems to be simply wrong. 

I strongly suggest re-opening this issue and removing the false check, so we can look for an appropriate solution for the original issue.

jgrandja:
@hbergmey Please see this [comment](https://github.com/spring-projects/spring-security-oauth/commit/0f732e48f2ec967756003758c6134bb1230671e2#commitcomment-27726611). Does this make sense?

PR #1147, opened 2017-08-29
Fix IllegalArgumentException on malformed JOSE header

vpavic:
This PR addresses the `IllegalArgumentException` that's thrown when `JwtHeaderConverter#convert` is invoked with JWT that contains a malformed JOSE header. This exception causes status `500` to be returned to the client instead of the expected `401`.

The PR includes a test that can be used to reproduce the problem.

This affects `master`, `2.1.x` and `2.0.x`.

On a sidenote, `JwtHeaderConverter#convert` javadoc claims that `JwkException` is thrown if the JWT is invalid while in reality it throws an `InvalidTokenException`. Can I use this PR to address that as well?


jgrandja:
Thanks for the PR @vpavic. I also updated the javadoc for `throws InvalidTokenException`. This is merged to master.

Issue #1155, opened 2017-09-09
Scopes not presented for user's approval page when using JWT


adolfoweloy:
I have experienced that when using JWT, the scopes isn't being presented to the user's approval page. I am adding JWT support by declaring a `JwtTokenStore` bean as follows.

```java
@Bean
public JwtTokenStore tokenStore() {
    return new JwtTokenStore(accessTokenConverter());
}
``` 
When a tokenStore is defined, a `TokenStoreUserApprovalHandler` is created and it does not propagate scopes from request to the respective model attributes. Maybe it should be similar to what happens inside `ApprovalStoreUserApprovalHandler.getUserApprovalRequest` because even when working with JWT there should be useful to present required scopes on user's approval page.

I saw that when `AuthorizationServerEndpointsConfigurer` is configured with `approvalStoreDisabled`, a `TokenStoreUserApprovalHandler` is defined and it's expected to not return scopes in the approval model, so it constrained me to simply add scopes to the approval model inside `TokenStoreUserApprovalHandler`.

I would like to know if it could be considered an issue that deserves some Pull Request.

jgrandja:
@adolfoweloy Please see [comment](https://github.com/spring-projects/spring-security-oauth/pull/1157#pullrequestreview-408146641)

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

spring-issuemaster:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

PR #1157, opened 2017-09-09
Add scopes to user approval request for TokenStoreUserApprovalHandler

adolfoweloy:
This pull request addresses the issue #1155 so the Authorization Server can present scopes on user's approval page even when using `JwtTokenStore` to support JWT tokens.

### DIFF
--- a/spring-security-oauth2/src/test/java/org/springframework/security/oauth2/provider/approval/TokenStoreUserApprovalHandlerTests.java
+++ b/spring-security-oauth2/src/test/java/org/springframework/security/oauth2/provider/approval/TokenStoreUserApprovalHandlerTests.java
@@ -12,10 +12,6 @@
  */
 package org.springframework.security.oauth2.provider.approval;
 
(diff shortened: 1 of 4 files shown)


spring-security-oauth2/src/test/java/org/springframework/security/oauth2/provider/approval/TokenStoreUserApprovalHandlerTests.java
```diff
@@ -75,6 +78,21 @@ public void testMemorizedApproval() {
 		assertTrue(handler.isApproved(authorizationRequest, userAuthentication));
 	}
 
+	@Test
+	public void testScopedApproval() {
```
jgrandja, line 82: It's not clear to me whether this is an issue or not. Can you please add a test that demonstrates the issue without this change taking effect.

jgrandja:
Closing due to lack of requested feedback

Issue #1159, opened 2017-09-15
Add support for EC based JWK keys


duergner:
The `spring-security-jwt` project currently does not offer any support for Elliptic Curves based signatures. This could be added  by an `EllipticCurveVerifier` similar to the existing `RsaVerifier`. JCE has support for EC based signatures as well.

Once support for EC based signatures is added to `spring-security-jwt` `spring-security-oauth2` could be extended to read EC based keys from a JWK source. This would require adding a new `EcJwkDefinition` class and mainly extend `JwkDefinitionSource` as well as `JwkSetConverter` to also handle EC keys beside the already supported RSA keys.

duergner:
I created a first draft implementation for the extension to `spring-security-jwt` at: https://github.com/duergner/spring-security-oauth/commit/d12020a31c38677c0001b588adafff5741fed448 

A first draft for potential changes to `spring-security-oauth2` can be found at https://github.com/duergner/spring-security-oauth/commit/d7835a999e12dd89a2fa29fba66f2910684d9629

duergner:
@rwinch I opened the two PRs to deliver this feature. Would be great to get feedback on what should be done in addition to get it merged.

I wasn't really sure about what type of tests you would expect to be in place for this e.g.

jgrandja:
@duergner Thank you for submitting both PR's.

As an FYI, this project is in maintenance mode as our efforts our focused in the new OAuth/OIDC support (and JWT/JOSE) in Spring Security 5.0. Therefore, we are only accepting bug fixes and minor enhancements at this time.

I did take a quick look at both PR's and at first glance looks like it will be a clean merge. However, this will have to wait until 2.3 is released because of the new artifacts. I'm not sure when 2.3 will be released as of now as all our cycles are going into Spring Security 5.0 which is planned to be released on Nov 6. If I get any spare cycles before than I will take a look at the PR's in detail.

In the meantime, please add tests to both PR's.

Thanks for your efforts!

adolfoweloy:
@jgrandja, @duergner 
have you noticed that travis checking are breaking because of an environment variable?

jgrandja:
@adolfoweloy @duergner The Travis CI broken build issue is resolved via 11f24d6470d5241d9d167e0fce6756190e66170e

duergner:
@jgrandja I pushed an update to both PRs that should contain tests covering the new features. Would be great to get some feedback whether it is sufficient now to be ready for merging.

[...]

PR #1161, opened 2017-09-18
Add Elliptic Curve verification support for spring-security-jwt

duergner:
The implementation is inspired by the RSAVerifier and uses code from org.apache.xml.security.algorithms.implementations.SignatureECDSA to transcode from JWK to JCE compatible signatures.

It is meant to be used by the patch to `spring-security-oauth2` which introduces support for *Elliptic Curve* based *JWK* keys.

This PR is one of the two PRs to deliver #1159

### DIFF
--- a/spring-security-jwt/src/main/java/org/springframework/security/jwt/crypto/sign/EllipticCurveVerifier.java
+++ b/spring-security-jwt/src/main/java/org/springframework/security/jwt/crypto/sign/EllipticCurveVerifier.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2006-2011 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package org.springframework.security.jwt.crypto.sign;
+
+import java.math.BigInteger;
+import java.security.GeneralSecurityException;
+import java.security.Signature;
+import java.security.interfaces.ECPublicKey;
+
+/**
+ * @author Michael Duergner
+ */
+public class EllipticCurveVerifier implements SignatureVerifier {
+
+    private final ECPublicKey key;
+
+    private final String algorithm;
+
+    public EllipticCurveVerifier(final BigInteger x, final BigInteger y, final String curve, final String algorithm) {
+        this(EllipticCurveKeyHelper.createPublicKey(x, y, curve), algorithm);
+    }
+
+    public EllipticCurveVerifier(final ECPublicKey key, final String algorithm) {
+        this.key = key;
+        this.algorithm = algorithm;
+    }
+
+    @Override
+    public String algorithm() {
+        return algorithm;
+    }
+
+    @Override
+    public void verify(byte[] content, byte[] sig) {
+        try {
+            Signature signature = Signature.getInstance(algorithm);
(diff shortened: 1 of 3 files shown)


spring-security-jwt/src/test/java/org/springframework/security/jwt/crypto/cipher/EllipticCurveKeyData.java
```diff
@@ -0,0 +1,34 @@
...
+package org.springframework.security.jwt.crypto.cipher;
+
+import org.springframework.security.jwt.codec.Codecs;
+
+import java.math.BigInteger;
+
+/**
+ * @author Michael Duergner
+ */
+public class EllipticCurveKeyData {
```
jgrandja: Please remove this class and put test data directly in `EllipticCurveVerificationTests`

spring-security-jwt/src/test/java/org/springframework/security/jwt/crypto/sign/EllipticCurveVerificationTests.java
```diff
@@ -0,0 +1,61 @@
...
+    byte[] content = Codecs.utf8Encode("eyJhbGciOiJFUzI1NiJ9.eyJrZXkiOiJ2YWx1ZSJ9");
+
+    EllipticCurveVerifier verifier = new EllipticCurveVerifier(
+        EllipticCurveKeyData.X,
+        EllipticCurveKeyData.Y,
+        EllipticCurveKeyData.CRV,
+        EllipticCurveKeyData.ALG);
+    verifier.verify(content, sig);
+  }
+}
```
jgrandja: Please add a failed `verify` test

spring-security-jwt/src/main/java/org/springframework/security/jwt/crypto/sign/EllipticCurveVerifier.java
```diff
@@ -0,0 +1,133 @@
...
+ * PLEASE NOTE: This implementation expects the signature to be supplied in the format as defined
+ * by the JWS spec. It will take care of transcoding it to DER format as expected by JCE.
+ *
+ * @author Michael Duergner
+ */
+public class EllipticCurveVerifier implements SignatureVerifier {
+
+    private final ECPublicKey key;
+
+    private final String algorithm;
```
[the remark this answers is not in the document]
duergner: As it is only a verifier at the moment I do not see any real benefit in doing so as the exact data needs to be read. If you see benefits in having a default algorithm I would suggest going with `SHA256withECDSA` (similar to the RSA version)

duergner:
@jgrandja I addressed the comments you had. Regarding the default algorithm I don't have any preference.

jgrandja:
Thanks @duergner. I'll take a look at this shortly.

spring-security-jwt/src/main/java/org/springframework/security/jwt/crypto/sign/EllipticCurveVerifier.java
```diff
@@ -0,0 +1,124 @@
...
+            if (!signature.verify(transcodeSignatureToDER(sig))) {
+                throw new InvalidSignatureException("EC Signature did not match content");
+            }
+        }
+        catch (GeneralSecurityException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private byte[] transcodeSignatureToDER(byte[] jwsSignature) throws GeneralSecurityException {
```
jgrandja: @duergner I'm circling back to this PR now. Looking at this bit of code I'm wondering if this is necessary? JCE and the provider we are using, Bouncy Castle, provides full support for EC. Is this conversion necessary before verification? I'd rather avoid this and let the provider handle the conversion.

duergner: I was not able to the not converted signatures verified - but I'm happy to remove this piece of conversion in case you have an idea on how to offload it to JCE/BouncyCastle.

jgrandja:
Thanks for the PR @duergner! I added some polish in a separate commit. This is now in master.

PR #1162, opened 2017-09-18
Add support for Elliptic Curve JWK signatures

duergner:
This will utilize the changes to spring-security-jwt for handling Elliptic Curve signatures. It integrates fully into the existing codebase and will simply allow to verify signatures based on Elliptic Curve when they are read from a JWK definition.

This PR depends on #1161 to be merged. Once this PR is merged a new version of `spring-security-jwt` should be released and this PR needs to be updated to use the final version instead of the `1.0.9.BUILD-SNAPSHOT` version it currently uses.

This PR is the seconds part of delivering #1159

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkAttributes.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkAttributes.java
@@ -19,6 +19,7 @@ package org.springframework.security.oauth2.provider.token.store.jwk;
  * Shared attribute values used by {@link JwkTokenStore} and associated collaborators.
  *
  * @author Joe Grandja
+ * @author Michael Duergner
  */
 final class JwkAttributes {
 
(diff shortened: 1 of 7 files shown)


spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkAttributes.java
```diff
@@ -55,8 +56,23 @@
...
+	 */
+	static final String EC_PUBLIC_KEY_X = "x";
+
+	/**
+	 * The &quot;y&quot; parameter contains the y value for a EC public key.
+	 */
+	static final String EC_PUBLIC_KEY_Y = "y";
+
+	/**
+	 * The &quot;crv&quot; (curve) parameter contains the urve value for a EC public key.
```
jgrandja: Typo -> urve

duergner:
@jgrandja I addressed the review comments. Would be great if you could have another look.

duergner:
This branch would still need one small update for referencing the release version of `spring-security-jwt` once it has been release as it is referencing the **BUILD-SNAPSHOT** version at the moment.

jgrandja:
Thanks for the PR @duergner. I added some polish in a separate commit and merged this to master.

PR #1181, opened 2017-10-04
TokenEndpoint: Logging generic exception with ERROR level

jmisur:
I need to suppress client-side errors in a certain log appender, but right now everything is logged with INFO level.
All other exception handlers are handling client-side errors, only this one is handling generic unexpected problems (e.g. NPE in downstream service). 
Because it's most likely server-side implementation problem and not client request problem, it should log with ERROR level.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/TokenEndpoint.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/TokenEndpoint.java
@@ -163,7 +163,7 @@ public class TokenEndpoint extends AbstractEndpoint {
 	
 	@ExceptionHandler(Exception.class)
 	public ResponseEntity<OAuth2Exception> handleException(Exception e) throws Exception {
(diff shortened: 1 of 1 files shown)


jgrandja:
Resolved via #1232

jmisur:
Hi @jgrandja. Thanks for looking at the problem. However I'm not sure the fix in https://github.com/spring-projects/spring-security-oauth/commit/0bc2a67ff6ce5069652d7a5412e61e0f147ff78e solves it properly. Now also for e.g. ClientAlreadyExistsException there will be error message in the logs, but this error is clearly a client side problem (return code 401). There is no way of distinguishing Internal Server Errors (e.g. wrong service impl) from client errors, which will happen all the time.

jgrandja:
@jmisur I'm not sure I follow you when you say `ClientAlreadyExistsException`...is clearly a client side problem? `ClientAlreadyExistsException` is thrown by `ClientRegistrationService.addClientDetails` which is server-side.

Just to clarify, this commit only addresses the fact that when an `Exception` occurs, than `logger.error` should be called **NOT** `logger.info`. It's not trying to address any other issue as you have pointed out.

PR #1209, opened 2017-11-15
Fix OAuth2AuthenticationDetails display string

dennyverbeeck:
The logic for inserting comma's between the components of the string was incorrect and would produce strings like e.g. `remoteAddress=0:0:0:0:0:0:0:1, tokenType=BearertokenValue=<TOKEN>` (missing comma between `tokenType` and `tokenValue` fields). This fix correctly inserts commas before appending a new field if there was already a field added before it.


jgrandja:
Thanks for the PR @dennyverbeeck. This is now in master.

Issue #1211, opened 2017-11-15
Unknown Exception handling is hard to trace


xenoterracide:
instead of getting a stacktrace errors are simply logged as info, and given the message, this is an expecial problem with the generic exception handling of `TokenEndopoint.handleException`

related to #1200 and #1202

jgrandja:
Duplicate of #1200

Issue #1200, opened 2017-10-26
Vulnerability: Information Exposure through exception


xenoterracide:
using oauth2:2.0.14

The reason for the error is bad code on my part, but the exception isn't handled properly, and is hard to debug

https://cwe.mitre.org/data/definitions/209.html

this is the response from the server
```
{"error":"server_error","error_description":"PreparedStatementCallback; uncategorized SQLException for SQL [insert into oauth_access_token (token_id, token, authentication_id, user_name, client_id, authentication, refresh_token) values (?, ?, ?, ?, ?, ?, ?)]; SQL state [72000]; error code [1461]; ORA-01461: can bind a LONG value only for insert into a LONG column\n; nested exception is java.sql.SQLException: ORA-01461: can bind a LONG value only for insert into a LONG column\n"}
```

here's the logs

```
[INFO ] [11:53:28] [http-nio-8080-exec-7] endpoint.TokenEndpoint:166 - (:O88trBJRSemI92Lu_CuC3Q) - Handling error: UncategorizedSQLException, PreparedStatementCallback; uncategorized SQLException for SQL [insert into oauth_access_token (token_id, token, authentication_id, user_name, client_id, authentication, refresh_token) values (?, ?, ?, ?, ?, ?, ?)]; SQL state [72000]; error code [1461]; ORA-01461: can bind a LONG value only for insert into a LONG column_; nested exception is java.sql.SQLException: ORA-01461: can bind a LONG value only for insert into a LONG column_
```

which is the only indication that this is from TokenEndpoint:166

in order to address the security risk I suggest replacing the response with a generic InternalServerException

in order to make this Exception more debuggable, I would suggest changing the logger to do the following which will allow slf4j to print the stack trace to the logs, I might suggest logging as error too.

```
"handling error: {}", e.getClass().getSimpleName(), e)
```

xenoterracide:
Bump, so can we do something about this?

jgrandja:
@xenoterracide Would you be interested in submitting a PR for this? Resources are tight right now as we are currently focused in building the new OAuth support in Spring Security 5.0.

xenoterracide:
[...]

PR #1212, opened 2017-11-15
Better Unknown Exception Handling

xenoterracide:
better handle unknown exceptions per #1200 and #1211, addresses the security vulnerability, and would provide needed information in the log to debug.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/error/DefaultWebResponseExceptionTranslator.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/error/DefaultWebResponseExceptionTranslator.java
@@ -66,8 +66,10 @@ public class DefaultWebResponseExceptionTranslator implements WebResponseExcepti
 			return handleOAuth2Exception(new MethodNotAllowed(ase.getMessage(), ase));
 		}
 
-		return handleOAuth2Exception(new ServerErrorException(e.getMessage(), e));
-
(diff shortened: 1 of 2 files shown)


spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/error/DefaultWebResponseExceptionTranslator.java
```diff
@@ -111,8 +113,8 @@ public int getHttpErrorCode() {
 	@SuppressWarnings("serial")
 	private static class ServerErrorException extends OAuth2Exception {
 
-		public ServerErrorException(String msg, Throwable t) {
```
jgrandja: No need to remove this constructor. Let's keep it. This needs to be a passive and binary compatible change.

xenoterracide: given that the class itself is private is it necessary for binary compatibility? but sure I can make that change

xenoterracide: also worth saying my constructor isn't public for any particular reason, just copying the code around it

jgrandja:
Closing as per [comment](https://github.com/spring-projects/spring-security-oauth/issues/1200#issuecomment-345847497)

PR #1223, opened 2017-11-20
Ignore unsupported JWK types when parsing JWK set

### LINKED ISSUE #1213: JWK Set parsing fails if provider's set contains non-RSA keys
Our OIDC Provider supports multiple JWS algorithms for signing access tokens and ID tokens, and therefore its `jwks_uri` provides both RSA and EC keys. For example:

```json
{
  "keys": [
    {
      "kty": "RSA",
      "e": "AQAB",
      "use": "sig",
      "kid": "B9ch",
      "n": "..."
    },
    {
      "kty": "EC",
      "use": "sig",
      "crv": "P-256",
      "kid": "J3DJ",
      "x": "...",
      "y": "..."
    },
    {
      "kty": "EC",
      "use": "sig",
      "crv": "P-384",
      "kid": "uNgL",
      "x": "...",
      "y": "..."
    },
    {
      "kty": "EC",
      "use": "sig",
      "crv": "P-521",
      "kid": "YW0I",
      "x": "...",
      "y": "..."
    }
  ]
}
```

Unfortunately, Spring Security OAuth is unable to handle such JWK Set since `JwkSetConverter#createJwkDefinition` throws `JwkException` the moment it runs into a non-RSA key, regardless of the fact that JWK Set actually does contain an RSA key.

IMO the behavior of the JWK Set parsing should be changed to ignore the keys of unsupported types, until #1159 gets merged, which is currently targeted at the `2.3.0` release. This affects `2.0.x`/`2.1.x`/`2.2.x`.

I could put together a PR to address this in time for `2.2.1` release if we agree on the general approach on the changes to be made.

vpavic:
At present, `JwkSetConverter` throws a `JwkException` when it attempts to parse a JWK with unsupported key type. This causes problems with providers that support multiple JWS algorithms, and therefore publish JWK set with keys of different types.

This PR updates the behavior of `JwkSetConverter` to ignore JWKs with unsupported key type.

This resolves #1213

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkSetConverter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkSetConverter.java
@@ -45,6 +45,7 @@ import static org.springframework.security.oauth2.provider.token.store.jwk.JwkAt
  * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7517#page-10">JWK Set Format</a>
  *
  * @author Joe Grandja
+ * @author Vedran Pavic
  */
 class JwkSetConverter implements Converter<InputStream, Set<JwkDefinition>> {
 	private final JsonFactory factory = new JsonFactory();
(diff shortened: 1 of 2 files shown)


jgrandja:
Thanks for the PR @vpavic! This is now merged to master.

PR #1282, opened 2018-02-02
Modify schema.sql LONGVARBINARY type to BLOB type.

SystemOutPrint:


Rainy99:
Brother, thank you so much.

tests/annotation/jdbc/src/main/resources/schema.sql
```diff
@@ -25,29 +25,29 @@ create table oauth_client_details (
 
 create table oauth_client_token (
   token_id VARCHAR(256),
-  token LONGVARBINARY,
+  token BLOB,
```
jgrandja, line 28: @SystemOutPrint It's not clear to me why this change is being proposed. There is no description added. The build passes with these tests so I'm going to close this PR. Feel free to submit a new PR with details on the proposed changes.

PR #1291, opened 2018-02-21
Make client id and scope attribute names of the converted token configurable

### LINKED ISSUE #1214: `JwtAccessTokenConverter` uses non-standard `client_id` and `scope` claims
`JwtAccessTokenConverter` attempts to extract information about client ID and scope from non-standard claims named `client_id` and `scope`, respectively. This behavior is actually encapsulated in `DefaultAccessTokenConverter` to which `JwtAccessTokenConverter` delegates to, and the constants used for claim names are defined in `AccessTokenConverter` interface.

Our OIDC Provider aligns with the [OAuth 2.0 Token Exchange (draft-ietf-oauth-token-exchange-09)](https://tools.ietf.org/html/draft-ietf-oauth-token-exchange-09) which defines the following JWT claims:

- ["scp" (Scopes) Claim](https://tools.ietf.org/html/draft-ietf-oauth-token-exchange-09#section-4.2)
- ["cid" (Client Identifier) Claim](https://tools.ietf.org/html/draft-ietf-oauth-token-exchange-09#section-4.3)

It would be nice if `JwtAccessTokenConverter` would allow to specify the claims to be used to obtain client ID and scope information, since the currently used ones are non-standard.

vpavic:
At present, `DefaultAccessTokenConverter` uses `client_id` and `scope` attribute names in the converted token, which translates to claim names when JWT based access tokens are used. This causes problems with providers that align with OAuth 2.0 Token Exchange (https://tools.ietf.org/html/draft-ietf-oauth-token-exchange-12) and use `cid` and `scp` claims.

This PR updates the `DefaultAccessTokenConverter` to make client id and scope attribute names of the converted token configurable.

This resolves #1214

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
@@ -33,13 +33,17 @@ import org.springframework.security.oauth2.provider.OAuth2Request;
  * Default implementation of {@link AccessTokenConverter}.
  * 
  * @author Dave Syer
- * 
+ * @author Vedran Pavic
  */
 public class DefaultAccessTokenConverter implements AccessTokenConverter {
 
 	private UserAuthenticationConverter userTokenConverter = new DefaultUserAuthenticationConverter();
 	
 	private boolean includeGrantType;
+
+	private String scopeAttribute = SCOPE;
+
+	private String clientIdAttribute = CLIENT_ID;
 
 	/**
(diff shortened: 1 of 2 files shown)


jgrandja:
Thanks for the PR @vpavic! This is now in master.

cremersstijn:
@vpavic how can you overwrite the "scope" property? Does it needs to be done in the properties file?

vpavic:
@cremersstijn I'm not sure what exactly do you mean by _overwrite the "scope" property_. If you're referring to replacing the default attribute used to resolve scope, you need to register your own `DefaultAccessTokenConverter` bean, set the desired `scopeAttribute` and wire it to your resource server configuration.

PR #1324, opened 2018-03-26
Provide more extension points to DefaultAccessTokenConverter

david-noel:
This pull request makes it easier for developers to subclass and override parts of the class.

1.  Getters added to member vars so that they can be accessed in a sublass
2.  Convert private methods to protected to allow them to be overridden
3.  Added a protected extractRequestAuthorities method to allow subclasses to customize the non-user authorities in the token.

Item 3 allows a developer to include authorities like ROLE_ADMIN in a JWT but have the AccessTokenConverter add individual permission authorities(CAN_DO_X, CAN_DO_Y) based on what the role has authority to do.   In my use case the ROLE_ADMIN has 100+ individual permissions and I do not want to include them all in the JWT.  They are statically defined for the roles and I want to add them while decoding the token.  Without this change I would have to cut and paste this whole file into my code base.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
@@ -75,7 +75,7 @@ public class DefaultAccessTokenConverter implements AccessTokenConverter {
 	/**
 	 * Flag to indicate that the grant type should be included in the converted token.
 	 *
-	 * @return the flag value (default false)
+	 * @return the flag value
 	 */
 	public boolean isIncludeGrantType() {
 		return includeGrantType;
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
@@ -55,12 +55,30 @@ public class DefaultAccessTokenConverter implements AccessTokenConverter {
 	}
 
 	/**
-	 * Flag to indicate the the grant type should be included in the converted token.
+	 * Converter for the part of the data in the token representing a user.
+	 *
+	 * @return the set userTokenConverter
+	 */
+	public UserAuthenticationConverter getUserTokenConverter() {
+		return userTokenConverter;
+	}
+
+	/**
+	 * Flag to indicate that the grant type should be included in the converted token.
 	 * 
 	 * @param includeGrantType the flag value (default false)
 	 */
 	public void setIncludeGrantType(boolean includeGrantType) {
 		this.includeGrantType = includeGrantType;	
+	}
+
+	/**
+	 * Flag to indicate that the grant type should be included in the converted token.
+	 *
+	 * @return the flag value (default false)
+	 */
+	public boolean isIncludeGrantType() {
+		return includeGrantType;
 	}
 
 	/**
@@ -74,6 +92,15 @@ public class DefaultAccessTokenConverter implements AccessTokenConverter {
 	}
 
 	/**
+	 * Scope attribute name to be used in the converted token.
+	 *
+	 * @return the scope attribute name to use
+	 */
+	public String getScopeAttribute() {
+		return scopeAttribute;
+	}
+
+	/**
 	 * Set client id attribute name to be used in the converted token. Defaults to
 	 * {@link AccessTokenConverter#CLIENT_ID}.
 	 *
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@david-noel Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1324)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1324) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@david-noel Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1324)!

david-noel:
Any thoughts on this PR?  It has been sitting around for a while.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
```diff
@@ -55,14 +55,32 @@ public void setUserTokenConverter(UserAuthenticationConverter userTokenConverter
 	}
 
 	/**
-	 * Flag to indicate the the grant type should be included in the converted token.
+	 * Converter for the part of the data in the token representing a user.
+	 *
+	 * @return the set userTokenConverter
+	 */
+	public UserAuthenticationConverter getUserTokenConverter() {
```
jgrandja: I don't feel this getter is required. Is there a specific use case where you need access to this?

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
```diff
@@ -55,14 +55,32 @@ public void setUserTokenConverter(UserAuthenticationConverter userTokenConverter
...
 	public void setIncludeGrantType(boolean includeGrantType) {
 		this.includeGrantType = includeGrantType;	
 	}
 
+	/**
+	 * Flag to indicate that the grant type should be included in the converted token.
+	 *
+	 * @return the flag value
+	 */
+	public boolean isIncludeGrantType() {
```
jgrandja: This does not need to be exposed. It's only required internally for `convertAccessToken` and `extractAuthentication`.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
```diff
@@ -73,6 +91,15 @@ public void setScopeAttribute(String scopeAttribute) {
 		this.scopeAttribute = scopeAttribute;
 	}
 
+	/**
+	 * Scope attribute name to be used in the converted token.
+	 *
+	 * @return the scope attribute name to use
+	 */
+	public String getScopeAttribute() {
```
jgrandja: This does not need to be exposed. It's only required internally for `convertAccessToken` and `extractAuthentication`.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
```diff
@@ -83,6 +110,15 @@ public void setClientIdAttribute(String clientIdAttribute) {
 		this.clientIdAttribute = clientIdAttribute;
 	}
 
+	/**
+	 * Client id attribute name to be used in the converted token.
+	 *
+	 * @return the client id attribute name to use
+	 */
+	public String getClientIdAttribute() {
```
jgrandja: Same comments apply as per `getScopeAttribute()`

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
```diff
@@ -147,21 +183,25 @@ public OAuth2Authentication extractAuthentication(Map<String, ?> map) {
...
+	protected Collection<GrantedAuthority> extractRequestAuthorities(Map<String, ?> map) {
+		if (map.containsKey(AUTHORITIES)) {
+			@SuppressWarnings("unchecked")
+			String[] roles = ((Collection<String>) map.get(AUTHORITIES)).toArray(new String[0]);
+			return AuthorityUtils.createAuthorityList(roles);
+		}
+		return null;
+	}
+
+	protected Collection<String> extractAudience(Map<String, ?> map) {
```
jgrandja: We cannot rename `getAudience` to `extractAudience` in a patch or minor release. This can only happen in a major release as we follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html). Also, we do not plan on delivering a 3.0 release. I'm not sure you are aware but please see this [announcement](https://spring.io/blog/2018/01/30/next-generation-oauth-2-0-support-with-spring-security).

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultAccessTokenConverter.java
david-noel: The getters were added so that child classes could access the private members (userTokenConverter, includeGrantType, scopeAttribute, clientIdAttribute) from overriden methods such as convertAccessToken and extractAuthentication.  Changing the scope of these new getters to protected or changing the fields from private to protected would also work.

Issue #784, opened 2016-06-15
Add support for 'CERTIFICATE', 'PKCS7' and 'PKCS12' types in RsaKeyHelper


rnavarropiris:
The [RsaKeyHelper](https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-jwt/src/main/java/org/springframework/security/jwt/crypto/sign/RsaKeyHelper.java) class (used for public/private key based signature verification) does not allow configuring PEM certificates with the formats:

```
-----BEGIN CERTIFICATE-----
...
-----END CERTIFICATE-----
```

```
-----BEGIN PKCS7-----
...
-----END PKCS7-----
```

```
-----BEGIN PKCS12-----
...
-----END PKCS12-----
```

PR #1338, opened 2018-04-17
Add support for reading public key from X.509 PEM certificates

imgx64:
Fixes (partially) #784


jgrandja:
Thanks for the PR @imgx64. This is now merged via 9ece323.

Issue #1347, opened 2018-04-27
"unknown (use) is currently not supported" when PublicKeyUse is not part of JWK


senkum:
According to the JWK Spec https://tools.ietf.org/html/rfc7517#section-4.2, PublicKeyUse is optional, but currently spring-security-oauth2 code throws exception when PublicKeyUse is not specified in the JWK

Yasinjama:
I am running into this issue as well. @senkum have you found a workaround until this get's fixed?

jgrandja:
Closing based on [comment](https://github.com/spring-projects/spring-security-oauth/pull/1348#discussion_r191760386)

PR #1348, opened 2018-04-27
Made PublicUseKey to be optional according to the JWK Spec

senkum:
Fix for issue  https://github.com/spring-projects/spring-security-oauth/issues/1347" 

unknown (use) is currently not supported" when PublicKeyUse is not part of JWK

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkSetConverter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkSetConverter.java
@@ -189,11 +189,13 @@ class JwkSetConverter implements Converter<InputStream, Set<JwkDefinition>> {
 		}
 
 		// use
-		JwkDefinition.PublicKeyUse publicKeyUse =
-				JwkDefinition.PublicKeyUse.fromValue(attributes.get(PUBLIC_KEY_USE));
-		if (!JwkDefinition.PublicKeyUse.SIG.equals(publicKeyUse)) {
-			throw new JwkException((publicKeyUse != null ? publicKeyUse.value() : "unknown") +
-					" (" + PUBLIC_KEY_USE + ") is currently not supported.");
+		JwkDefinition.PublicKeyUse publicKeyUse = null;
+		if (attributes.containsKey(PUBLIC_KEY_USE)) {
+			publicKeyUse = JwkDefinition.PublicKeyUse.fromValue(attributes.get(PUBLIC_KEY_USE));
+			if (!JwkDefinition.PublicKeyUse.SIG.equals(publicKeyUse)) {
+				throw new JwkException((publicKeyUse != null ? publicKeyUse.value() : "unknown") +
+						" (" + PUBLIC_KEY_USE + ") is currently not supported.");
+			}
 		}
 
 		// alg
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkSetConverter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkSetConverter.java
@@ -137,13 +137,15 @@ class JwkSetConverter implements Converter<InputStream, Set<JwkDefinition>> {
 		}
 
 		// use
-		JwkDefinition.PublicKeyUse publicKeyUse =
-				JwkDefinition.PublicKeyUse.fromValue(attributes.get(PUBLIC_KEY_USE));
-		if (!JwkDefinition.PublicKeyUse.SIG.equals(publicKeyUse)) {
-			throw new JwkException((publicKeyUse != null ? publicKeyUse.value() : "unknown") +
-					" (" + PUBLIC_KEY_USE + ") is currently not supported.");
+		JwkDefinition.PublicKeyUse publicKeyUse = null;
+		if (attributes.containsKey(PUBLIC_KEY_USE)) {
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@senkum Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1348)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1348) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@senkum Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1348)!

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkSetConverter.java
```diff
@@ -137,13 +137,15 @@ private JwkDefinition createRsaJwkDefinition(Map<String, String> attributes) {
 		}
 
 		// use
-		JwkDefinition.PublicKeyUse publicKeyUse =
-				JwkDefinition.PublicKeyUse.fromValue(attributes.get(PUBLIC_KEY_USE));
-		if (!JwkDefinition.PublicKeyUse.SIG.equals(publicKeyUse)) {
-			throw new JwkException((publicKeyUse != null ? publicKeyUse.value() : "unknown") +
-					" (" + PUBLIC_KEY_USE + ") is currently not supported.");
+		JwkDefinition.PublicKeyUse publicKeyUse = null;
+		if (attributes.containsKey(PUBLIC_KEY_USE)) {
```
Yasinjama: I see this check will bypass the validation, but will this default to "sig" or "enc" when validating the signature.

senkum: According to spec it's optional, they did not specify any default value. Also I went through the code the PublicKeyUse is not used any where in the code.

jgrandja: @senkum @Yasinjama Although the spec does say the `use` parameter is optional, it is required by `JwkTokenStore` because it only supports signature (`sig`)  verification and NOT payload decryption (`enc`).

If we make this optional in `JwkTokenStore` that how do we know if the JWK is meant for `sig` or `enc`?

senkum: @jgrandja, I went through the code base , PublicKeyUse is not used any where in the code. 

Can we default to sig when PublicKeyUse is not present?

jgrandja: @senkum PublicKeyUse is used during the validation of the JWK(s) after received from the JWK Set Endpoint. This is required as the JWK can be used for `sig` or `enc`. We cannot default to `sig`. If the returned JWK is meant for `enc` and we default to `sig` this would introduce a bug.

senkum: @jgrandja I may be wrong, but when I searched the entire code base, 

The only use of the PublicKeyUse i found are in the following places
   - JwkSetConverter in the methods createRsaJwkDefinition and createEllipticCurveJwkDefinition, the validation is done, this is where exception is thrown when the value is null.
   - JwkSetConverterTest tests for JwkSetConverter
   - JwkDefinition - member variable, this variable is not used anywhere

Nowhere else it's used, please correct me if I am wrong

jgrandja: @senkum You are correct, `PublicKeyUse` is used/required in:

> JwkSetConverter in the methods createRsaJwkDefinition and createEllipticCurveJwkDefinition, the validation is done, this is where exception is thrown when the value is null.

As mentioned in my previous comment, this validation is required because it's possible that one of the JWK is meant for `enc`. Therefore, we need to distinguish if the JWK is meant for `sig` or `enc`. This is why we require the `use` parameter. We cannot default to `sig` for all returned JWK's as this would introduce a bug if the JWK is actually meant for `enc`.

jgrandja: @senkum I'm going to close this PR based on the reason I mentioned in my previous [comment](https://github.com/spring-projects/spring-security-oauth/pull/1348#discussion_r191760386).

sjoerd-michels: This should not have been closed: the implementation does not follow the spec, so this is a bug.
Now, for example, Spring security does not work with the following JWK key set that is perfectly valid but that does not specify a use attribute:
`{ "keys": [
    { "kty": "RSA",
      "x5t#S256": "...",
      "e": "...",
      "x5t": "...",
      "kid": "SIGNING_KEY",
      "x5c": [
        "...",
        "..."
      ],
      "key_ops": [
        "encrypt",
        "verify"
      ],
      "alg": "RS256",
      "n": "..."
    }
  ]
}`
And unfortunately, it seems impossible to adapt the `JwkSetConverter`.
This should be re-opened / re-considered.

preethamveeranna: which artifact version has got this fix?

PR #1351, opened 2018-04-29
Update DefaultRedirectResolver

kiyoung-Lee:
Hi.
I found setMatchSubdomains function is not used anywhere in DefaultRedirectResolver class and matchSubdomains boolean variable default value is true. So I'm updating hostMatches function. because I think hostMatchesfunction condition is always true.

I also update hostMatches access modifier to 'private'. bacause I think that function is only use in DefaultRedirectResolver class.

Thank you!

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/DefaultRedirectResolver.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/DefaultRedirectResolver.java
@@ -40,18 +40,7 @@ public class DefaultRedirectResolver implements RedirectResolver {
 
 	private Collection<String> redirectGrantTypes = Arrays.asList("implicit", "authorization_code");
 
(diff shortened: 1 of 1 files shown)


spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/DefaultRedirectResolver.java
```diff
@@ -40,19 +40,8 @@
...
-	private boolean matchSubdomains = true;
-
 	private boolean matchPorts = true;
 
-	/**
-	 * Flag to indicate that requested URIs will match if they are a subdomain of the registered value.
-	 * 
-	 * @param matchSubdomains the flag value to set (deafult true)
-	 */
-	public void setMatchSubdomains(boolean matchSubdomains) {
```
jgrandja, line 52: `setMatchSubdomains()` cannot be removed as it can break existing applications that are using it.

PR #1397, opened 2018-06-07
Make AuthorizationCodeTokenGranter follow rfc6749

acktsap:
According to [rfc 6749](https://tools.ietf.org/html/rfc6749#section-4.1.3), redirect_uri is required on access token request.
And i add a simple test case for token request with no redirect_uri.


acktsap:
I misunderstood

Issue #1410, opened 2018-06-23
RefreshTokenGranter should handle AuthenticationException


aschatten:
_Similar to gh-1408_

1. Generate a valid access and refresh token for a user.
2. Deactivate user.
3. Access `oauth` endpoint trying to refresh token.
4.`RefreshTokenGranter` does NOT handle `AccountStatusException`(or `UsernameNotFoundException`) exceptions properly, it bubbles up to the API level and results into `401` response status. This behaviour is confusing, because for the token endpoint `401` means that client authentication failed. 

Expected results:
At step 3, `RefreshTokenGranter` must convert `AccountStatusException` and `UsernameNotFoundException` to `InvalidGrantException`, similar to what `ResourceOwnerPasswordTokenGranter` does to `AccountStatusException` and `BadCredentialsException`. If the exception if converted to `InvalidGrantException`, then it will set status code to `400`.

PR #1411, opened 2018-06-23
RefreshTokenGranter handles AuthenticationException

aschatten:
Fixes gh-1410


jgrandja:
Thanks for the PR @aschatten. This is now in master.

Issue #1355, opened 2018-05-03
Getting invalid token using spring boot autoconfigure for the resource server


brsvpr:
Hi, 
I am trying to secure my Rest APIs. 
I have created an auth server which is a spring boot project running separately.
I have put the auth server as below. 

public class AuthServerOAuth2Config extends AuthorizationServerConfigurerAdapter {
	
	private static final Logger log = LoggerFactory.getLogger(AuthServerOAuth2Config.class);
	
	@Autowired
	private AuthenticationManager authenticationManager;
	
	@Override
	public void configure(ClientDetailsServiceConfigurer clients) throws Exception {
		clients
			.inMemory()
			.withClient("test")
			.secret("testsecret")
			.authorizedGrantTypes("client_credentials")
			.scopes("write", "read");
	}	
	
	@Override
    public void configure(AuthorizationServerSecurityConfigurer security) throws Exception {
        /*
         * Allow our tokens to be delivered from our token access point as well as for tokens
         * to be validated from this point
         */
        security.checkTokenAccess("permitAll()");
    }

	@Override
	public void configure(AuthorizationServerEndpointsConfigurer endpoints) throws Exception {
		endpoints.tokenStore(new InMemoryTokenStore()).authenticationManager(authenticationManager);
	}
}

So auth server code is working and I am able to access a token by using the below end point url.
http://localhost:9092/uaa/oauth/token

I am able to validate the token / get the client info by using the below end point url.
http://localhost:9092/uaa/oauth/check_token?token=08ed1c9c-a371-43cc-b456-407fcde0b315

I have created a resource server which is a spring boot project running separately which has few Rest APIs accessible by end point URLs that needs to be secured. Below are the properties in the application.properties.
security.oauth2.resource.token-info-uri=http://localhost:9092/uaa/oauth/check_token
security.oauth2.resource.token-type=Bearer
security.oauth2.client.client-id=test
security.oauth2.client.client-secret=testsecret
security.oauth2.client.grant-type=client_credentials
security.oauth2.client.scope=write

I would assume that by mentioning the above properties, spring boot will first validate the token against the auth server by calling the token-info-uri mentioned above.

Now I am testing with POSTMAN by passing the "Bearer <TOKEN>" as Authorization header to the one of the APIs in the resource server. I am getting the below error.
{
    "error": "invalid_token",
    "error_description": "08ed1c9c-a371-43cc-b456-407fcde0b315"
}

[...]

PR #1440, opened 2018-07-30
remote-check-token-invalid

Clement-Leopold:
issue #1355 Getting invalid token using spring boot autoconfigure for the resource server 

when I try to use the default RemoteTokenServices to check whether my token is valid in Spring Cloud, it failed. 
And When I debug line by line,  I find the line 116 in RemoteTokenServices checking the 'active' whether that is true or not, it failed to tell when it is true.  
So I override the RemoteTokenServices, and changed the line to be this, it succeeded.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
@@ -12,6 +12,7 @@
  *******************************************************************************/
 package org.springframework.security.oauth2.provider.token;
 
+import com.sun.org.apache.xpath.internal.operations.Bool;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.http.HttpEntity;
@@ -41,7 +42,7 @@ import java.util.Map;
  *
  * @author Dave Syer
  * @author Luke Taylor
- *
+ * @author Xianhao Chen
  */
 public class RemoteTokenServices implements ResourceServerTokenServices {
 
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@Clement-Leopold Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1440)!

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
```diff
@@ -113,7 +113,7 @@ public OAuth2Authentication loadAuthentication(String accessToken) throws Authen
 		}
 
 		// gh-838
-		if (!Boolean.TRUE.equals(map.get("active"))) {
+		if (!Boolean.TRUE.toString().equals(map.get("active"))) {
```
jgrandja: @Clement-Leopold With this change, `"true"` would work, but `true` would not. The provider may return the `active` attribute as either a JSON boolean or string. We need to account for both types.

Can you please update and add a test as well?

Clement-Leopold:
I've added one test and modified the condition code, if you have any suggestions please let me know, thanks!

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/RemoteTokenServices.java
Clement-Leopold: @jgrandja  I've made some updates as you requested. thanks!

Clement-Leopold:
@jgrandja  I am so sorry for that once I've pushed the author name.  This is my fault.

jgrandja:
@Clement-Leopold I haven't had a chance to review this as of yet. We've been very busy working towards the upcoming Spring Security 5.1 release. I'm planning on circling back to this within the next couple of weeks. Thanks for your patience.

jgrandja:
@Clement-Leopold I'm going to close this PR as it's been addresses (and merged) in #1533

PR #1442, opened 2018-08-01
RFC 7662 Token introspection username parameter

elniko:
RFC 7662 specifies the response parameters for token introspection and it contains username in place of user_name.
With username parameter, it will be possible to use DefaultUserAuthenticationConverter when you use RemoteTokenServices and an introspect link to a third party authorization server to validate a token

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/UserAuthenticationConverter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/UserAuthenticationConverter.java
@@ -27,7 +27,7 @@ public interface UserAuthenticationConverter {
 
 	final String AUTHORITIES = AccessTokenConverter.AUTHORITIES;
 
-	final String USERNAME = "user_name";
+	final String USERNAME = "username";
 
 	/**
 	 * Extract information about the user to be used in an access token (i.e. for resource servers).


Kaszaq:
To add on this - as using `user_name` is valid per RFC 7662 if considered as extension to that RFC, most services implementing that RFC follow proposed OPTIONAL name for username which is `username`, hence it might be hard to integrate such services with Spring which expects `user_name` https://tools.ietf.org/html/rfc7662#section-2.2

bencody:
I also noticed this today. In my opinion, this is a bug. The specification doesn't mention `user_name` at all, just `username`.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/UserAuthenticationConverter.java
```diff
@@ -27,7 +27,7 @@
 
 	final String AUTHORITIES = AccessTokenConverter.AUTHORITIES;
 
-	final String USERNAME = "user_name";
+	final String USERNAME = "username";
```
jgrandja, line 30: Changing this will break existing applications.

jgrandja:
Even though RFC 7662 specifies `username` it is still OPTIONAL. And from my experience, the user name may be represented as another completely different claim other than `username` or `user_name` depending on the provider implementation.

Closing this PR in favour of #1696 since it provides greater flexibility.

PR #1456, opened 2018-08-13
Make ClientDetails return Collection<? extends GrantedAuthority> type

### LINKED ISSUE #1186: ClientDetails should Collection<? extends GrantedAuthority> getAuthorities() like UserDetails
I am facing a bug while implementing my client.

We have to use a ClientDetails for authentication, this interface discribed here:

https://github.com/spring-projects/spring-security-oauth/blob/0f732e48f2ec967756003758c6134bb1230671e2/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/ClientDetails.java#L82

needs to have an object type `Collection<GrantedAuthority>` instead of `Collection<? extends GrantedAuthority>` like in **[UserDetails](https://docs.spring.io/spring-security/site/docs/current/apidocs/org/springframework/security/core/userdetails/UserDetails.html)**:

https://github.com/spring-projects/spring-security/blob/master/core/src/main/java/org/springframework/security/core/userdetails/UserDetails.java#L54

This result in dumb code to convert the list, is it possible to patch it ?

Jamesxu182:
fixed #1186

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/ClientDetails.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/ClientDetails.java
@@ -79,7 +79,7 @@ public interface ClientDetails extends Serializable {
 	 * 
 	 * @return the authorities (never <code>null</code>)
 	 */
-	Collection<GrantedAuthority> getAuthorities();
+	Collection<? extends GrantedAuthority> getAuthorities();
 
 	/**
 	 * The access token validity period for this client. Null if not set explicitly (implementations might use that fact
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@Jamesxu182 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1456)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1456) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Jamesxu182 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1456)!

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/ClientDetails.java
```diff
@@ -79,7 +79,7 @@
 	 * 
 	 * @return the authorities (never <code>null</code>)
 	 */
-	Collection<GrantedAuthority> getAuthorities();
+	Collection<? extends GrantedAuthority> getAuthorities();
```
jgrandja, line 82: This change cannot be applied since it will break existing applications.

PR #1478, opened 2018-09-10
Support clock skew for access token expiry check

### LINKED ISSUE #1287: Add support for clock skew when checking access token expiry
By standard JWT deliver an expiration date with resolution in seconds. This is used to set the field ``expiration`` of type ``java.util.Date`` in DefaultOAuth2AccessToken.

The DefaultOAuth2AccessToken (2.2.1.RELEASE) methods in question are realized as follows:
```
	public boolean isExpired() {
		return expiration != null && expiration.before(new Date());
	}

	public int getExpiresIn() {
		return expiration != null ? Long.valueOf((expiration.getTime() - System.currentTimeMillis()) / 1000L)
				.intValue() : 0;
	}
```

As ``new Date()`` generates a current time representation in milliseconds and this is not truncated to second values, the calculation of "isExpired" returns ``false`` for 999ms longer, than getExpiresIn() returns 0.

Example: Given expiration is at 10s = 10000ms and currentTimeMillis is 9001ms.
getExpiresIn = (10000 - 9001)/1000 = 999/1000 = 0
isExpired = 10000 before 9001 = 10000 < 9001 = false

Expected: isExpired() == (getExpiresIn() == 0)

This actually leads to real bug behavior. For example: ``org.springframework.security.oauth2.client.OAuth2RestTemplate#getAccessToken()`` uses ``OAuth2AccessToken#isExpired()`` to decide, whether to acquire a fresh token or to use the existing token. For almost one second, this will definitly lead to a an unauthorized request, if the server side actually validates seconds correctly. This is a behavior we actually observe in communication between a Spring-based client and a WSO2 secured server.

### LINKED ISSUE #1477: OAuth2RestTemplate does not account for network latency / clock skew
When making a request via OAuth2RestTemplate (In our case we are using client credentials), The getAccessToken method checks if the token is expired and if so, requests a new token:

```Java
OAuth2AccessToken accessToken = context.getAccessToken();

if (accessToken == null || accessToken.isExpired()) {
try {
	accessToken = acquireAccessToken(context);
}
```

If the token is 1ms from expiration then this code will still attempt to use the token. The server will reject this token when it receives it (This is also a problem if the server/client have small difference in clocks).

It seems like the best-practice would be to refresh the token ahead of the actual expiration by some small grace period (i.e. 5 seconds).

kPOWz:
expiration. Fixes #1287,  fixes #1477.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/OAuth2RestTemplate.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/OAuth2RestTemplate.java
@@ -50,6 +50,8 @@ public class OAuth2RestTemplate extends RestTemplate implements OAuth2RestOperat
 
 	private OAuth2RequestAuthenticator authenticator = new DefaultOAuth2RequestAuthenticator();
 
+	private int tokenExpirationDelta = 5;
+
 	public OAuth2RestTemplate(OAuth2ProtectedResourceDetails resource) {
 		this(resource, new DefaultOAuth2ClientContext());
 	}
@@ -157,6 +159,11 @@ public class OAuth2RestTemplate extends RestTemplate implements OAuth2RestOperat
 		return resource.getClientId();
 	}
 
+	private boolean isApproachingExpiration(OAuth2AccessToken accessToken) {
+		int expiresIn = accessToken.getExpiresIn();
+		return accessToken.isExpired() || (expiresIn != 0 && expiresIn <= this.tokenExpirationDelta);
+	}
+
 	/**
 	 * Acquire or renew an access token for the current context if necessary. This method will be called automatically
 	 * when a request is executed (and the result is cached), but can also be called as a standalone method to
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@kPOWz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1478)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1478) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@kPOWz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1478)!

btrummer:
How are the odds that this pull request gets included and released soon (within this month)?
Thanks.

jgrandja:
@btrummer This PR adds a new public method `setTokenExpirationDelta()`, which can only be added in a minor version - see [Semantic versioning](https://semver.org/spec/v2.0.0.html) for more info.

We don't plan on releasing a new minor version anytime soon. I'm not sure you are aware but this project is in maintenance mode and we're focusing all our efforts in the new OAuth support in Spring Security 5.x. We will fix bugs but limiting adding new features. Please see this [post](https://spring.io/blog/2018/01/30/next-generation-oauth-2-0-support-with-spring-security) for more context.

jgrandja:
Thanks for the PR @kPOWz. FYI, I added some polish before merging to master.

PR #1489, opened 2018-09-25
fix accessToken to ThreadLocal

bigbaldy1128:
accessToken field should be a thread local variable,because the getMap function which in org.springframework.boot.autoconfigure.security.oauth2.resource.UserInfoTokenServices may be called by multiple threads at the same time. this problem had been reproduced in my performance test environment.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/DefaultOAuth2ClientContext.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/DefaultOAuth2ClientContext.java
@@ -10,51 +10,51 @@ import org.springframework.security.oauth2.common.OAuth2AccessToken;
 
 /**
  * The OAuth 2 security context (for a specific user or client or combination thereof).
- * 
+ *
  * @author Dave Syer
  */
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@bigbaldy1128 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1489)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1489) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@bigbaldy1128 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1489)!

jgrandja:
It's not clear to me why `DefaultOAuth2ClientContext.accessToken` should be stored in a `ThreadLocal`. At minimum a test would need to be provided that reproduces the issue.

Also, this PR is stale and needs to be rebased on master (2.4.x). If this is still valid, please rebase and submit a new PR.

PR #1519, opened 2018-11-06
Refactor Token Endpoint

kenricksatrio:
**Changes**

* Currently there is a OAuth2Utils class that has constant of common token request's parameter key.
Instead of using literal string, use of this constant value.
`"grant_types"` → `OAuth2Utils.GRANT_TYPE`
`"scopes"` → `OAuth2Utils.SCOPE`
* Combine unnecessary nested if


jgrandja:
Thanks for the PR @kenricksatrio. This is now in master.

Issue #941, opened 2017-01-06
NullPointerException if refresh_token not set


codemaker219:
I did a mistake in one of my clients and forget to set the refresh_token.
I use oauth2 within spring-boot version 1.4.2.RELEASE. I try to get an access-token with a refresh-token but do not set the refresh token:
Request:
---> POST http://XXX/oauth/token HTTP/1.1
Content-Type: application/x-www-form-urlencoded
Accept: application/json
Content-Length: 24
Authorization: Basic XXX

grant_type=refresh_token

Response:
<--- HTTP/1.1 500 (78ms)
Transfer-Encoding: chunked
X-Frame-Options: DENY
Cache-Control: no-store
X-Content-Type-Options: nosniff
Connection: close
Pragma: no-cache
X-XSS-Protection: 1; mode=block
X-Application-Context: XXX
Date: Fri, 06 Jan 2017 12:36:41 GMT
Content-Type: application/json;charset=UTF-8

{"error":"server_error","error_description":null}

I think it would be better if the code would be 400 with the description “refresh token not present” or similar.  That would have helped me.

rnavarropiris:
I had the same issue.
The problem seems to rely in the RefreshTokenGranter class, since it is not checked if the value of the refresh_token is not set. See: https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/refresh/RefreshTokenGranter.java#L46

PR #1542, opened 2018-12-14
Fix NPE when refresh_token is not provided

whiteship:
Related issue: #941 

The NullPointerException can occure when the `grant_type` is `refersh_token` but the `refresh_token` itself is not provided. And the response code would be 500 as it is a server error.

With this fix, it will return 4xx as it is one of the client side error that missed providing required parameter, `refresh_token`.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/TokenEndpoint.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/TokenEndpoint.java
@@ -129,6 +129,9 @@ public class TokenEndpoint extends AbstractEndpoint {
 		}
 
 		if (isRefreshTokenRequest(parameters)) {
+			if (parameters.get("refresh_token") == null) {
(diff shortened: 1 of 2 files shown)


spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
```diff
@@ -137,6 +139,11 @@ public OAuth2AccessToken refreshAccessToken(String refreshTokenValue, TokenReque
 			throw new InvalidGrantException("Invalid refresh token: " + refreshTokenValue);
 		}
 
+		if (refreshTokenValue == null) {
```
jgrandja, line 142: The `refresh_token` validation should happen before it reaches `DefaultTokenServices`. We should move it to either `TokenEndpoint` or `RefreshTokenGranter`.

jgrandja:
Thanks for the PR @whiteship. This is now in master.

whiteship:
Thank you @jgrandja.

PR #1543, opened 2018-12-14
Extracting Bearer token from multi value header

micd:
When an `Authorization` header consists of multiple token types e.g. `Basic auth, Bearer auth`, `extractHeaderToken` will return null whereas it would be good to receive `auth`. I'm not sure if anyone uses authorization header like that, but it might be possible. 

I have prepared PR to extract token correctly in such case. Check out if it's needed or not.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/authentication/BearerTokenExtractor.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/authentication/BearerTokenExtractor.java
@@ -13,81 +13,87 @@
 
 package org.springframework.security.oauth2.provider.authentication;
 
-import java.util.Enumeration;
-
-import javax.servlet.http.HttpServletRequest;
-
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@micd Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1543)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1543) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@micd Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1543)!

jgrandja:
Thanks for the PR @micd. However, I'm reluctant on merging this.

When an OAuth Client calls a Resource Server then it should be `Bearer` token. If a `Basic` token is sent in addition to the `Bearer` token then this should really be fixed in the client.

Based on this explanation, I'm going to close this PR.

Issue #1545, opened 2018-12-18
Log generic exceptions in TokenEndpoint to ERROR


jbiblio:
I know that logging of Exceptions in `TokenEndpoint` (and maybe other parts of the lib) are discussed controversially (e.g. #1290 , #1232).

The current state of 2.3.4 is:
- some Exceptions are logged to WARN log, `HttpRequestMethodNotSupportedException` to INFO log
- all stack traces are omitted because of
`logger.info("Handling error: " + e.getClass().getSimpleName() + ", " + e.getMessage());` and
`logger.warn("Handling error: " + e.getClass().getSimpleName() + ", " + e.getMessage());`

IMHO that's fine for Exceptions defined by the protocol itself, e.g. a `ClientRegistrationException` or bad credentials.

On the other hand, this produces log lines on production stages like the following:
`2018-12-18 10:56:57,432 WARN  (ajp task-3:) [org.springframework.security.oauth2.provider.endpoint.TokenEndpoint] Handling error: NullPointerException, null`

There's no chance to debug such NPEs without patching the application (e.g. by applying an Aspect as suggested in https://github.com/spring-projects/spring-security-oauth/issues/1290#issuecomment-372543662.

I'd like to suggest to log `Excepetion` with it's stack trace in `org.springframework.security.oauth2.provider.endpoint.TokenEndpoint.handleException(Exception)`, since they represent unhandled Exceptions. Some specialised Exceptions are handled by the Framework or Protocol (e.g. by returning a 401 to the client), so omitting the stack trace seems to be OK.

Another possibility is to distinguish between WARN/INFO and DEBUG:
`if (log.isDebugEnabled()) log.debug(e.getMessage(), e); else log.warn(e.getMessage());`
That way it would be possible to enable stack traces without patching the product

TIA!

jgrandja:
@jbiblio Thank you for the feedback. Would you be interested in submitting a PR for this improvement?

jbiblio:
@jgrandja Sure, I'll take a look.

PR #1555, opened 2019-01-12
Log generic Exceptions to ERROR with stack trace

jbiblio:
As described in #1545, generic Exceptions like NullPointerExceptions are written to WARN log with stack trace omitted, so it's impossible to trace the root cause of such Exceptions.

This change attaches the Exception to the logger, so its full stack trace will be logged. Additionally, the log level will be ERROR.

All other Exceptions related to the protocol, like Bad Credentials, are treated in separate ExceptionsHandlers, so we assume that every other Exception is unexpected and should get some attention.


pivotal-issuemaster:
@jbiblio Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1555)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1555) to manually synchronize the status of this Pull Request.

jgrandja:
Thanks for the PR @jbiblio. This is now in master via ec78c7b

PR #1573, opened 2019-02-04
Making oauth2 compatible between jdbc and jwtconverter.

mplescano:
I want to merge these changes in the current master. I have implemented an oauth server using the jdbc storage along side jwt converter. I found a lot of issues and finally landed in this pull request.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/AccessTokenConverter.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/AccessTokenConverter.java
@@ -38,8 +38,12 @@ public interface AccessTokenConverter {
 	final String JTI = "jti";
 	
 	final String GRANT_TYPE = "grant_type";
+	
+	final String TOKEN_TYPE = "token_type";
 
 	final String ATI = "ati";
+	
+	final String RTI = "rti";
 
 	final String SCOPE = OAuth2AccessToken.SCOPE;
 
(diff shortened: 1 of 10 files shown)


jgrandja:
@mplescano Thanks for the PR. However, it's not clear to me if this PR is addressing a specific bug?

> I found a lot of issues and finally landed in this pull request

You don't mention any existing issues that this PR is attempting to fix.

I will need more information on why you are proposing these changes. Also, there are failing tests that need to be resolved as well.

mplescano:
Hello @jgrandja

ok let me explain:

* The principal struggle I had found was in refresh token, the refresh token had a dependency into access token id (by refreshTokenInfo.put(ACCESS_TOKEN_ID, tokenId)). 
This is not scalable when I want to generate a new access token from the refresh token. that's because I eliminated the dependency or relationship putting refreshTokenInfo.put(TOKEN_TYPE, "refresh_token") instead.

* In the access token I put this info.put(REFRESH_TOKEN_ID, encodedRefreshToken.getValue()) in order to reverse the relationship, I mean to know the refresh token id from the access token.

This PR is for integrating both jdbc and JWT token in the same base code.

I don't know hot to resolve the failing tests because these are occurring beyond the code I touched.

Regards

jgrandja:
@mplescano The enhancement applied in this PR which eliminates the dependency the refresh token has with the access token will break existing applications. There were quite a few existing tests that were changed in order this to pass, which is another indication that this will break existing applications.

I understand that the association doesn't make sense from your standpoint:

> The principal struggle I had found was in refresh token, the refresh token had a dependency...

It doesn't stop you from applying this change on your end. All you need to do is configure a custom `JwtAccessTokenConverter` (or `TokenEnhancer`).

Given that the enhancement is non-passive, we won't be able to merge this.

PR #1626, opened 2019-03-16
Address Bias in RandomValueStringGenerator

### LINKED ISSUE #639: Bias in RandomValueStringGenerator Authorization Code Generation
The transformation of random bytes to alphanumeric characters using the modulo operator in the RandomValueStringGenerator#getAuthorizationCodeString() method introduces a bias such that the first 8 elements in the character array are 25% more likely to be selected than the other elements.

This is because the length of DEFAULT_CODEC is 62, which does not evenly divide into 256. The bias prevents the the random value strings from being uniformly random, which may allow an attacker to recover nonrandom, predictable values using brute force methods.

A simple fix is to add two more characters to DEFAULT_CODEC, bringing its length to 64. Recommend adding URL safe characters hyphen (-) and underscore (_).

jokerr:
Fixes #639.

I added the URL safe characters hyphen (`-`) and underscore (`_`) as recommended.  If there are other characters that are preferred let me know.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/RandomValueStringGenerator.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/RandomValueStringGenerator.java
@@ -1,3 +1,15 @@
+/*
+ * Copyright 2011-2019 the original author or authors.
+ *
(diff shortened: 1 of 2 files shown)


jgrandja:
Thanks for the PR @jokerr. This is now in master.

PR #1660, opened 2019-04-12
RedisTokenStore expiration fix #1657

### LINKED ISSUE #1657: RedisTokenStore uname_to_access and client_id_to_access leak
### Summary

When the access key is stored, there is being set several expirations on keys related to the access key.

Unfortunately there are two redis keys that are not expired correctly. These keys are with prefixes **_client_id_to_access:_** and **_uname_to_access:_**. In following example I am going to describe only the **_client_id_to_access:_** keys.

### Actual Behavior

Lets imagine there is just one client - **web** - that is used by all the users that logs in to the application. Every time there is a new access token generated it is stored into the list (or set in newer RedisTokenStore implementation) **_client_id_to_access:web_**. The expiration is set to whole key (not just to new value added).

In case the access token expiration is 5 minutes and your app is regularly visited - there is no 5 minutes window that no one generates new access token - then this list (or set) gets bigger and bigger, containing all the tokens that were stored previously and are already expired from all the other keys (e.g. **_access:_** and so on).

### Version

We're using version 2.0.16 that is somehow older. On the other hand the same issue is even in the master branch. The only main difference between 2.0.16 and master branch is in used data structure. 2.0.16 uses the list (rPush operations) and master uses set (sAdd) that has the O(1) complexity to store and remove. But the expiration is completely the same.

LukasVyhlidka:

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/RedisTokenStore.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/RedisTokenStore.java
@@ -2,6 +2,7 @@ package org.springframework.security.oauth2.provider.token.store.redis;
 
 import org.springframework.data.redis.connection.RedisConnection;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
+import org.springframework.data.redis.connection.RedisZSetCommands;
 import org.springframework.data.redis.core.Cursor;
 import org.springframework.data.redis.core.ScanOptions;
 import org.springframework.security.oauth2.common.ExpiringOAuth2RefreshToken;
@@ -30,11 +31,10 @@ public class RedisTokenStore implements TokenStore {
 	private static final String AUTH_TO_ACCESS = "auth_to_access:";
 	private static final String AUTH = "auth:";
 	private static final String REFRESH_AUTH = "refresh_auth:";
-	private static final String ACCESS_TO_REFRESH = "access_to_refresh:";
 	private static final String REFRESH = "refresh:";
 	private static final String REFRESH_TO_ACCESS = "refresh_to_access:";
-	private static final String CLIENT_ID_TO_ACCESS = "client_id_to_access:";
-	private static final String UNAME_TO_ACCESS = "uname_to_access:";
+	private static final String CLIENT_ID_TO_ACCESS = "client_id_to_access_z:";
+	private static final String UNAME_TO_ACCESS = "uname_to_access_z:";
 
 	private static final boolean springDataRedis_2_0 = ClassUtils.isPresent(
 			"org.springframework.data.redis.connection.RedisStandaloneConfiguration",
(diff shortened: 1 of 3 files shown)


LukasVyhlidka:
The problem in this PR is backward compatibility. It has already happened that RedisTokenStore changed the data structure from list to set. This time it is changed from set to sorted set. The prefix could be changed to something little different (e.g. client_id_to_access_z:) - what do you think?
Or there can be created a bash script that could be used to transform the set data structure to sorted set.

LukasVyhlidka:
Changed the prefix based on previous comment. At our project I would like to deploy the new version without the downtime so it has to be "backward compatible"

OrangeDog:
There is still a problem with they key's expiration being set to the last-issued expiration time.
This isn't necessarily the latest expiration time, and indeed the set could already contain tokens that never expire.

OrangeDog:
While we're here, there doesn't seem to be any need for the `CLIENT_ID_TO_ACCESS` keys, because they are a prefix of the `UNAME_TO_ACCESS` keys.

LukasVyhlidka:
> There is still a problem with they key's expiration being set to the last-issued expiration time.
> This isn't necessarily the latest expiration time, and indeed the set could already contain tokens that never expire.

You're right. I left it because it was previous behavior and did not solve an issue I was experiencing. Because the cleaning is made on the sorted set (by score values), the actual expire on whole key could be probably removed, right? On the other hand, my implementation needs to be configured so the clean method is called once upon a time. In case it is not cleaned, there would be no cleaning at all.




> While we're here, there doesn't seem to be any need for the `CLIENT_ID_TO_ACCESS` keys, because they are a prefix of the `UNAME_TO_ACCESS` keys.
I don't understand what you mean. Uname is username and clientId is ID of the 3rd party app that authorizes the user (without knowing password). The client_id_to_access is needed in case I want to revoke all the tokens created for this 3rd party client app - e.g. because its password leaked or something else.

OrangeDog:
> Uname is username

Despite the name, it's actually the client id and the username, used for `findTokensByClientIdAndUserName`.

LukasVyhlidka:
> when the set is empty

Well, in case the empty set is problem, then yes. Is it problem?

> findTokensByClientIdAndUserName

What if I want to list all the tokens without knowing the username? In case I want all the tokens only by clientId? I can probably use SCAN for that but this looks somehow more reasonable.

OrangeDog:
> Is it problem?

Well yes. You'd end up with thousands and thousands of keys that are just empty sets.

>  can probably use SCAN for that

Exactly. You already use SCAN in the cleanup. Using it for `findTokensByClientId` reduces the number of keys that need to be managed.

OrangeDog:
On further investigation, an empty set doesn't actually exist, so there's no issue there.

LukasVyhlidka:
On the other hand, what is actually not needed is the ACCESS_TO_REFRESH object. It is only written and deleted, never read. And the TTL (expire) is not set corretly - it is the ttl of refresh token, but should be the ttl of access token.
Going to delete whole key in this PR.

LukasVyhlidka:
@jgrandja Hello, would you mind to take a look into this PR, please?

jgrandja:
@LukasVyhlidka We've been quite busy leading up to the upcoming Spring Security 5.2.0 GA release this Sep 25. After the release is out, I have a few items (including this one) in this project I need to attend to. Thanks for your patience.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/RedisTokenStore.java
```diff
@@ -30,11 +31,10 @@
 	private static final String AUTH_TO_ACCESS = "auth_to_access:";
 	private static final String AUTH = "auth:";
 	private static final String REFRESH_AUTH = "refresh_auth:";
-	private static final String ACCESS_TO_REFRESH = "access_to_refresh:";
 	private static final String REFRESH = "refresh:";
 	private static final String REFRESH_TO_ACCESS = "refresh_to_access:";
-	private static final String CLIENT_ID_TO_ACCESS = "client_id_to_access:";
-	private static final String UNAME_TO_ACCESS = "uname_to_access:";
+	private static final String CLIENT_ID_TO_ACCESS = "client_id_to_access_z:";
```
jgrandja: Changing the Key ID for both `CLIENT_ID_TO_ACCESS` and `UNAME_TO_ACCESS` will leave existing keys orphaned after a user upgrades to this version, which will ultimately break existing applications. Is there a reason why the Key ID's need to be changed?

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/RedisTokenStore.java
```diff
@@ -398,24 +399,68 @@ private void removeAccessTokenUsingRefreshToken(String refreshToken) {
...
+	/**
+	 * Runs a maintenance of the RedisTokenStore.
+	 *
+	 * SortedSets UNAME_TO_ACCESS and CLIENT_ID_TO_ACCESS contains access tokens that can expire.
+	 * This expiration is set as a score of the Redis SortedSet data structure. Redis does not support expiration of items in a container data structure.
+	 * It supports only expiration of whole key. In case there is still new access tokens being stored into the RedisTokenStore before whole key gets expired,
+	 * the expiration is prolonged and the key is not effectively deleted. To do "garbage collection" this method should be called once upon a time.
+	 * @return how many items were removed
+	 */
+	public long doMaintenance() {
```
jgrandja: This cannot go into a patch release since it's `public`. Please see the [changes allowed](https://apr.apache.org/versioning.html#strategy) for a patch release.

Also, I don't think it makes sense to have a "maintenance" routine in the `TokenStore` implementation. I think this responsibility should be in a separate component altogether.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/RedisTokenStore.java
OrangeDog: If they're not changed then existing applications will also break, because they'll the wrong data types at those keys.

jgrandja:
@LukasVyhlidka Closing this PR since the feedback hasn't been addressed. If you are able to address the feedback provided, we can re-open and continue the review.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/RedisTokenStore.java
OrangeDog: How to maintain the token store depends on the internal details of the token store. I'm not sure a separate component would be a good design.

OrangeDog:
> In case there is a request to delete all the tokens for a client, the Scan could take quite a lot of time

The time taken for a `SCAN` depends on the number of keys in the db, not the number of keys that will be returned. The total time can be reduced by increasing the count argument, but that can block other operations for longer.

Issue #1193, opened 2017-10-25
JwtAccessTokenConverter will change the refreshToken even when reuseRefreshToken set to true


bwangfdu:
I find that the JwtAccessTokenConverter will change the refreshToken when reuseRefreshToken set to true. It happens on the line 247:
 DefaultOAuth2RefreshToken token = new DefaultOAuth2RefreshToken(
					encode(encodedRefreshToken, authentication));
after the encode function is called, the refreshToken was changed (re-encoded). But the DefaultTokenServices will not store the new one cause by Line 178: 
if (!reuseRefreshToken) {
			tokenStore.storeRefreshToken(accessToken.getRefreshToken(), authentication);
		}

To reproduce this case, you can set reuseRefreshToken to true and set the tokenEnhancer to JwtAccessTokenConverter, and after call refresh_token process, there will be no refresh_token can refresh anymore.

loesak:
I can confirm I am having the same issue.

The authorization server returns a new refresh token every time a refresh token is used regardless if "reuseRefreshToken" is set to "true" or not. The refresh token isn't being stored in the TokenStore. This is because the JwtAccessTokenConverter is creating a new refresh token when its "enhance" method is called.

Key Dependencies:
spring-cloud-starter-oauth2:1.2.3.RELEASE
spring-security-oauth2:2.0.15.RELEASE
spring-security-jwt:1.0.9.RELEASE
spring-cloud-starter-security:1.2.3.RELEASE

bwangfdu:
This issue was resolved in later version. It is caused by the order of scopes.

bwangfdu:
I am sorry, the current problem is much different from when I raise this issue two years ago.

czeroprime:
I can confirm this is still an issue also, to expand on what @loesak has said, the method 'JwtAccessTokenConverter.enhance' updates the 'ati' property with the token id('jti' property) from the new access token.

This update to the refresh token is not stored in the database in the 'DefaultTokenServices.refreshAccessToken' method due to the if statement 'if (!this.reuseRefreshToken) {' on line 123.

As the token does in fact change, should the if statement 'if (!this.reuseRefreshToken) {' be removed to fix this issue? And if so the if statement 'if (!this.reuseRefreshToken) {' on line 116 should be removed as this will remove the existing refresh token.

[...]

Issue #1109, opened 2017-07-06
setting reuseRefreshToken to true not working.


MadhuraMaddali:
Hi,
I am using JwtTokenStore and I am trying to reuse Refresh Token.

I set up the configuration to support reusing the refresh token in two places.
endpoints.reuseRefreshTokens(true); and 
tokenService.setReuseRefreshToken(true);
        
I am observing that when refreshToken call is made, JwtAccessTokenConverter enhance method is called and is returning new refreshToken irrespective of the settings.

Code from JwtAccessTokenConverter enhance method where it is returning the new refresh token.
{code}
DefaultOAuth2RefreshToken token = new DefaultOAuth2RefreshToken(encode(
					encodedRefreshToken, authentication));
			
{code}

In My understanding, reuseRefreshToken setting it to true should return the same refresh token.
Am I missing any thing? Is it known issue?

MadhuraMaddali:
Hi, 
Anyone facing similer issue ?

Jardo-51:
I noticed the same issue. 

`JwtAccessTokenConverter `creates a new refresh token and `DefaultTokenServices` just stores the old one (because reuse is turned on). The new refresh token is returned in response but doesn't work. In my understanding, the response should contain the old refresh token.

guedes91:
I am facing the same issuer. In my opinion It would not return the a new refresh token when you are using refresh_token grant type.

dfcoffin:
Section 6. Refreshing an Access Token of 'The OAuth 2.0 Authorization Framework' RFC [6749] states:

The authorization server MAY issue a new refresh token, in which case the client MUST discard to old refresh token and replace it with the new refresh token.  The authorization server MAY revoke the old refresh token after issuing a new refresh token to the client.  If a new refresh token is issued, the refresh token scope MUST be identical to that of the refresh token included by the client in the request.

Therefore, the issue is that the JwtAccessTokenConverter is complying with the standard, but the DefaultTokenServices is improperly saving the now revoked token.  Without reviewing the code, it would appear that the JwtAccessTokenConverter is NOT testing the reuse flag before completing the refresh token request.

jluo926:
Facing the same issue. Any update on this?

jureczkyb:
Facing the same issue.

1016990109:
Facing the same issue.

1016990109:
#1193

MrManiak:
[...]

PR #1668, opened 2019-04-29
Fix reuse refresh token flow

czeroprime:
Fixes: #1193 #1109 

This update fixes an issue with the refresh token flow when the 'reuseRefreshToken' flag is true.

Description of issue:
If the 'reuseRefreshToken' flag is true, and 'supportRefreshToken' flag is true, when refreshing the access token the refresh token changes also(it updates the 'ati' property with the id of the new access token). This new refresh token is referenced against the access token when stored, but the new refresh token itself is not stored, and the original refresh token not being linked to an existing access token. This results in refresh tokens being single use, and being orphaned after use. More information on the issue present in #1193 

Fix:
The fix removes the existing refresh token, and stores the new refresh token.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
@@ -168,16 +168,17 @@ public class DefaultTokenServices implements AuthorizationServerTokenServices, R
 
 		authentication = createRefreshedAuthentication(authentication, tokenRequest);
 
+		tokenStore.removeRefreshToken(refreshToken);
+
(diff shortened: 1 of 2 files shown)


RazmikMkrtchyan:
This way **reuseRefreshToken** not works properly. Every time generates new refresh token. I have resovled issued by removeing **encode** function from refresh token. Is it bad solution?

czeroprime:
Hi @RazmikMkrtchyan, @dcoffin pointed out in issue #1109 the spec does in fact say that a new refresh token can be returned on refresh.
I reference this and expand on it a little in issue #1193.

jgrandja:
Closing this PR. Please see this [comment](https://github.com/spring-projects/spring-security-oauth/issues/1109#issuecomment-546453794)

PR #1674, opened 2019-05-10
Remove use of deprecated APIs in samples

artem-smotrakov:
I wanted to play with the sample application but looks like they don't work:

* Fixed #1672 by replacing `ifAllGranted="..."` with `access="hasRole(...)"`
* `AbstractAuthenticationProcessingFilter.SPRING_SECURITY_LAST_EXCEPTION_KEY` has been [deprecated](https://docs.spring.io/spring-security/site/docs/3.1.x/apidocs/org/springframework/security/web/authentication/AbstractAuthenticationProcessingFilter.html#SPRING_SECURITY_LAST_EXCEPTION_KEY), replaced it with `WebAttributes.AUTHENTICATION_EXCEPTION`.

PR #1703, opened 2019-06-11
Safer deserialization in RedisTokenStore

### LINKED ISSUE #1682: Unsafe deserialization in RedisTokenStore
(after discussing on security@pivotal.io, it was decided to consider the below as a defense in depth improvement)

Making a long story short, compromising a Redis server may lead to compromising the whole application. It can happen due to unsafe deserialization. Of course, compromising a Redis server looks like a serious issue by itself. Redis servers should be protected and configured in a secure way. But nevertheless I am wondering if spring-security-oauth should be updated a bit to prevent arbitrary code execution in the application in case the Redis server is hacked.

Here are some details. An application can use `RedisTokenStore` which stores and reads auth info from a Redis server. `RedisTokenStore` uses the default Java serialization mechanism to prepare data to be stored in the Redis server. Here is how `ObjectInputStream` can be called:

```
  RedisTokenStore.getAccessToken()                       (spring-security-oauth)
    RedisTokenStore.deserializeAccessToken()             (spring-security-oauth)
      JdkSerializationStrategy.deserializeInternal()     (spring-security-oauth)
        JdkSerializationRedisSerializer.deserialize()    (spring-data-redis)
          DeserializingConverter.convert()               (spring-framework)
            DefaultDeserializer.deserialize()            (spring-framework)
              ConfigurableObjectInputStream.readObject() (spring-framework)
```

--- content deleted ---

You can see there are two pre-requisites to make this attack work:
- An attacker should be able to modify data in Redis.
- Application's classpath should contain classes to build a deserialization gadget.

The pre-requisites above may be quite hard to meet which makes it harder to successfully implement the attack. Nevertheless it may be better to update the framework and libs to mitigate it.

I have several patches which can address this issue. The patches update Spring Framework, Spring Data Redis and Spring Security OAuth. The main idea is to introduce whitelisting in `ConfigurableObjectInputStream` to be able to deserialize only allowed classes, and then, `JdkSerializationStrategy` can set a list of safe classes.

Another way to fix this is to use the filtering mechanism which was introduced in JEP 290 [1].
JEP 290 has been backported to JDK 6u141. So, Spring Framework would require at least JDK 6u141 then.

If the idea sounds okay, I can open pull requests, and we can finalize the solution.

[1] https://openjdk.java.net/jeps/290

artem-smotrakov:
Please review the following patch which fixes #1682 

- Added `SaferObjectInputStream` which checks if classes are allowed to deserialize.
- Updated `JdkSerializationStrategy` to use `SaferObjectInputStream` instead of `JdkSerializationRedisSerializer`.
- Updated `JdkSerializationStrategy` with a list of classes allowed to deserialize.
- Added a couple of tests for `JdkSerializationStrategy`

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/JdkSerializationStrategy.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/JdkSerializationStrategy.java
@@ -1,26 +1,83 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.springframework.security.oauth2.provider.token.store.redis;
 
-import org.springframework.data.redis.serializer.JdkSerializationRedisSerializer;
+import org.springframework.core.serializer.support.SerializationFailedException;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.ObjectOutputStream;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 
 /**
- * Serializes objects using {@link JdkSerializationRedisSerializer}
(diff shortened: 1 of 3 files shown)


artem-smotrakov:
Hi @jgrandja I know you're probably busy. When you get some time, could you please take a look at this patch and the ones below? They have been waiting for a review for several months. Thank you in advance!

#1706
#1674
#1708
#1711
#1717

jgrandja:
@artem-smotrakov Apologies for the delay. Been very busy leading up to the Spring Security 5.2.0 GA this Sep 25. After the release is out, I'll be circling back to this project to attend to a few items including all the ones you have listed. Thanks for your patience.

jgrandja:
Thank you for the PR @artem-smotrakov. This is now in master.
FYI, I made `SaferObjectInputStream` package-private since this is going into a patch release.

ldwqh0:
some times ,I have my owner UserDetails implementation in my package,
how can I serialize it to redis,
and how can i deserialize it from redis?

PR #1706, opened 2019-06-17
Don't log tokens in JdbcTokenStore

### LINKED ISSUE #1472: JdbcTokenStore always logs live tokens at INFO
JdbcTokenStore's logic for determining whether a token is present in the database is to issue a SELECT and catch (and log) an exception if it returns no results. This log message contains the token ID that was queried for, which is a password equivalent.

However, `storeAccessToken`, which is called at least on every token creation, uses this method and triggers that log message. This means that _every token issuance_ results in the live token's value being written to syslog at INFO level.

This is tangentially related to issue #1379.

artem-smotrakov:
Please review this simple patch which prevents logging tokens in `JdbcTokenStore`. The patch fixes #1472


jgrandja:
Thanks for the PR @artem-smotrakov. This is now in master.

Issue #1710, opened 2019-06-26
CheckTokenEndpoint should restrict to POST method only


sourabhsparkala:
### Summary
This is in relation to Issue #327

 `/oauth/check_token` endpoint accepts all HTTP methods. Instead, it should just accept `POST` method requests. If a client sends a request to the endpoint via GET request and provide a token in a URL, then the token may be written to the application's logs which is not a good security practice. 

### Actual Behavior

If a client sends a request to the endpoint via `GET` request and provides a token in a URL `/oauth/check_token?token={token}`

The `CheckTokenEndpoint` still accepts the request.
 
### Expected Behavior

If a client sends a request to the respective endpoint via `GET` or any another request (not `POST`), then it should return the response `405: METHOD_NOT_ALLOWED`

And

If a client sends a request to the respective endpoint via `POST` request, then it should return the response `200: OK`

The token parameter has to be part of the request body while sending the request.

By fixing this we can ensure that the users are always obligated to access oAuth endpoints with `POST` requests.

jlarroque:
is there any mechanism to temporarily support this? Even when is correct, perhaps it would be nice have a a backward compatibility mechanism to avoid having GET endpoint available

jgrandja:
@jlarroque 

> perhaps it would be nice have a a backward compatibility mechanism to avoid having GET endpoint available

The `CheckTokenEndpoint` only supports POST. It does not support GET

murilo8812:
> is there any mechanism to temporarily support this? Even when is correct, perhaps it would be nice have a a backward compatibility mechanism to avoid having GET endpoint available

to backward compatibiliy, only implement a new endpoint with the same name in any rest controller of your application. Example:

`@RequestMapping(value = "oauth/check_token", method = { RequestMethod.GET, RequestMethod.POST })
	public Map<String, ?> checkToken(@RequestParam("token") String value) {

		final DefaultAccessTokenConverter accessTokenConverter = new DefaultAccessTokenConverter();

		final OAuth2AccessToken token = resourceServerTokenServices.readAccessToken(value);

		if (token == null) {
			throw new InvalidTokenException("Token was not recognised");
		}

		if (token.isExpired()) {
			throw new InvalidTokenException("Token has expired");
		}

		OAuth2Authentication authentication = resourceServerTokenServices.loadAuthentication(token.getValue());

		@SuppressWarnings("unchecked")
[...]

PR #1711, opened 2019-06-26
Restricting OAuth to allow only POST method in check_token endpoint

sourabhsparkala:
Fixes: #1710

- Restricting `/oauth/check_token` in `CheckTokenEndpoint` to allow only `POST` requests
- Writing some unit testcases to make sure that the endpoint functionality does not cause regression issues


artem-smotrakov:
`continuous-integration/travis-ci/pr` failures doesn't seem to be caused by this patch. I hope it will succeed after rerun.

jgrandja:
Thanks for the PR @sourabhsparkala. This is now in master.

PR #1717, opened 2019-07-05
Redirect HTTP response code is set to 303 for Authorization Endpoint

### LINKED ISSUE #900: Support 303 status for authorization response
This issue specifically addresses the **[307 Redirect Attack](https://blog.acolyer.org/2016/11/07/a-comprehensive-formal-security-analysis-of-oauth-2-0/)**. A more detailed explanation can be found [here](https://arxiv.org/pdf/1601.01229v4.pdf).

In a nutshell, after the Authorization Server (IdP) successfully authenticates the user, typically through a HTTP POST of the user's credentials, the following HTTP response status (redirect) set by the Authorization Server **should not** set the HTTP status to a value that will instruct the user-agent to follow-up the request containing the user's credentials. This scenario would happen with a 307 status return after successful authentication.

The current implementation within Spring Security which Spring Security OAuth leverages is the `DefaultRedirectStrategy` which sets the status to 302. Although most modern browsers follow-up the 302 with a GET request and therefore doesn't expose the user's credentials in the request, there may be user-agents that implement the 302 with a follow-up POST instead of a GET which then would expose the user's credentials to the Relying Party on the callback.

In order to avoid the user's credentials being propagated on the callback request, ensure the Authorization Server sets the HTTP status to 303 on the authorization response.

sourabhsparkala:
Fixes #900 

- Default Redirect Strategy is set to HTTP Response Status Code 303. As per below-given conditions:
    * The Endpoint is receiving a POST request
    * The POST request is expecting a redirect
    * Handling every RedirectView present in AuthorizationEndpoint.java
- Modifying tests to handle regression issues

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/AuthorizationEndpoint.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/AuthorizationEndpoint.java
@@ -12,6 +12,7 @@
  */
 package org.springframework.security.oauth2.provider.endpoint;
 
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.AccessDeniedException;
 import org.springframework.security.authentication.InsufficientAuthenticationException;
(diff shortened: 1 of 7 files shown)


jgrandja:
Thanks again for the PR @sourabhsparkala. This is now in master.

PR #1718, opened 2019-07-05
Replace UUID with SecureRandom in  (DefaultTokenServices)

alxdrlitreev:
**Problem**
Usage of UUID for access tokens is unacceptable. 

**Why**
Accoding to [RFC 6749](https://tools.ietf.org/html/rfc6749#section-10.10) (OAuth 2.0 Authorization Framework Specs):

> The probability of an attacker guessing generated tokens (and other credentials not intended for handling by end-users) MUST be less than or equal to 2^(-128) and SHOULD be less than or equal to 2^(-160).

128 bits in a random-generated UUID v.4, there are 6 bits which are fixed variant and version bits leaving **only 122 bits of actual random which does not comply with RFC specifications.**

**Solution**
Using a secure source of random `SecureRandom`, we're generating base64 out of 160-bit random value.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
@@ -295,7 +295,7 @@ public class DefaultTokenServices implements AuthorizationServerTokenServices, R
 		SecureRandom random = new SecureRandom();
 		Base64.Encoder encoder = Base64.getUrlEncoder().withoutPadding();
 		byte[] accessTokenBuffer = new byte[20];
-		random.nextBytes(buffer);
+		random.nextBytes(accessTokenBuffer);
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@alxdrlitreev Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1718)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1718) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@alxdrlitreev Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1718)!

alxdrlitreev:
Last three commits fixing some code issues, I didn't

PR #1720, opened 2019-07-05
Replace UUID with SecureRandom in `createAccessToken` (DefaultTokenSe…

alxdrlitreev:
**Problem**
Usage of UUID for access tokens is unacceptable. 

**Why**
Accoding to [RFC 6749](https://tools.ietf.org/html/rfc6749#section-10.10) (OAuth 2.0 Authorization Framework Specs):

> The probability of an attacker guessing generated tokens (and other credentials not intended for handling by end-users) MUST be less than or equal to 2^(-128) and SHOULD be less than or equal to 2^(-160).

128 bits in a random-generated UUID v.4, there are 6 bits which are fixed variant and version bits leaving **only 122 bits of actual random which does not comply with RFC specifications.**

**Solution**
Using a secure source of random `SecureRandom`, we're generating base64 out of 160-bit random value.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
@@ -16,6 +16,9 @@ package org.springframework.security.oauth2.provider.token;
 import java.util.Date;
 import java.util.Set;
 import java.util.UUID;
+import java.security.SecureRandom;
+
+import org.apache.commons.codec.binary.Base64;
 
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.security.authentication.AuthenticationManager;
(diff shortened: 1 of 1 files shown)


jgrandja:
Thanks for the PR @alxdrlitreev. I added a bit of polish and just merged this to master.

bmaehr:
I think this change makes major problems. 

On my system there have different tokens been generated which contains unexpected characters and converting them back and forth there have been spaces and other url escaped characters. This is producing random problems at some requests for a user getting such a token.

alxdrlitreev:
@bmaehr it means you’re doing something wrong since random data is being encoded with base64 and cannot produce such issues as you’ve stated. 

If you’re still sure, that it’s not your fault (highly likely it is) than you have to provide sample of code, which causes such issues. 

Cheers!

bmaehr:
@alxdrlitreev Have a look at the ticket I opened, there is an example that produces problems.

Issue #1722, opened 2019-07-09
Thread contention issue when using bogus keyId


JanGroot:
### Actual Behavior

due to some misconfiguration on our platform, we’ve noticed the possibility to severely hamper a service by providing an unknown key in the header of a JWT token.

Our application uses Spring security and is configured with an org.springframework.security.oauth2.provider.token.store.jwk.JwkVerifyingJwtAccessTokenConverter

This class is used in the verification of all incoming requests.

If the “kid” value in the JWT token is unknown, the implementation  (JwkDefinitionSource) will block and retrieve the current JWK’s from the configured url. In our case this could take up to 500ms leading to contention. (application was under load of 100’s of requests per second a fraction of which used a token for a different environment, resulting in timeouts due to the blocking)

Issue is probably the clear() of the jwkdefenitions. Perhaps load the new definitions, put them and thn remove those not curent anymore.

### Expected Behavior

Ofcourse this is a tricky case, if there is a key rollover you'd like get the new keys. Mitigation could be in our config to cache the jwkseturl, but then you disallow possible ok sessions on rollover in the cache period.  Or do something with etag..

### Version
2.3.5

also present in the current version

JanGroot:
(I've mailed this issue a couple of months ago to security@pivotal and never heard back, forgot about it, and just now thought of it again, hence this more direct way)

PR #1723, opened 2019-07-10
clear jwk definitions *after* retrieving new ones

JanGroot:
An unknown jwk id triggers the retrieval of the key definitions as it usually means the keys rotated.

When this call executes the definition map was cleared before reading the current definitions
from the remote `jwkSetUrl`. 
During this retrieval subsequent requests can not be
verified, which is fine in case of a key rotation. But if someone intentionally changes the kid, other requests still have a valid key but are forced to block until the call completes.  

 Moving the clear and put after the retrieval of the new jwk
definitions mitigates this.

Closes gh-1722

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkDefinitionSource.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkDefinitionSource.java
@@ -96,10 +96,12 @@ class JwkDefinitionSource {
 			if (result != null) {
 				return result;
 			}
-			this.jwkDefinitions.clear();
+			Map newJwkDefinitions = new LinkedHashMap<String, JwkDefinitionHolder>();
 			for (URL jwkSetUrl : jwkSetUrls) {
-				this.jwkDefinitions.putAll(loadJwkDefinitions(jwkSetUrl));
+				newJwkDefinitions.putAll(loadJwkDefinitions(jwkSetUrl));
 			}
+			this.jwkDefinitions.clear();
+			this.jwkDefinitions.putAll(newJwkDefinitions);
 			return this.getDefinition(keyId);
 		}
 	}


spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkDefinitionSource.java
```diff
@@ -96,10 +96,12 @@ JwkDefinitionHolder getDefinitionLoadIfNecessary(String keyId) {
 			if (result != null) {
 				return result;
 			}
-			this.jwkDefinitions.clear();
+			Map newJwkDefinitions = new LinkedHashMap<String, JwkDefinitionHolder>();
 			for (URL jwkSetUrl : jwkSetUrls) {
-				this.jwkDefinitions.putAll(loadJwkDefinitions(jwkSetUrl));
+				newJwkDefinitions.putAll(loadJwkDefinitions(jwkSetUrl));
 			}
+			this.jwkDefinitions.clear();
```
jgrandja: @JanGroot Reorganizing the `clear()` just before the `putAll()` won't do anything to improve the performance here since `synchronized (this.jwkDefinitions)` will block until all JWK's have been fetched and stored. This will need to be re-worked in order to improve performance by reducing the blocking time.

JanGroot: Hi, thanks for looking into this.

The point of moving the clear to after the *loadJwkDefinitions* is to make sure the other threads with a correct KID never reach the synchronized block as they will still return on line 91 as the keys are not cleared yet and they will still be found.

The only threads waiting on the fetch will be those with an unknown or bogus key.

OLD situation:
definitions contains A
t1 enters with key B -> definitions is cleared -> start fetching keys
t2, t3, t4 , t5 enter with key A -> definition not found -> all block on fetch
t1 -> put keys (only A as B does not exist) -> returns null 
t2 returns key A

New 
definitions contains A
t1 enters with key B -> block on  start fetching keys
t2, t3, t4, t5, t6  enter with key A -> returns key A
t1 -> clears key -> puts keys -> returns null for B

JanGroot: Just seeing now this project is in maintenance, in spring-sec 5 this is ok already via
`com.nimbusds.jose.jwk.source.RemoteJWKSet#updateJWKSetFromURL`
there the cache is never cleared to begin with.

jgrandja: @JanGroot Yes, you are correct. Moving the `clear()`  and `putAll()` after the retrieval will help. I missed this in my initial analysis.

jgrandja:
Thanks for the PR @JanGroot. This is now in master.

Issue #959, opened 2017-01-24
Support fragment in userAuthorizationUri for redirecting user


hillmanli-seekers:
Some OAuth2 authorization provider contains fragment (the # anchor stuff) in their authorization uri, for example WeChat, (in this case WeChat redirect uri is https://open.weixin.qq.com/connect/qrconnect#wechat_redirect). Adding the fragment to userAuthorizationUri will cause OAuth2ClientContextFilter.redirectUser to throw IllegalArgumentException. It is due to using UriComponentsBuilder.fromHttpUrl.

I am able to workaround the problem by overriding OAuth2ClientContextFilter.redirectUser in a subclass, though it would be nice to have this out of the box.

jgrandja:
Closing this as fragment components are not allowed in `redirect_uri`. See this [comment](https://github.com/spring-projects/spring-security-oauth/pull/1724#issuecomment-607186136) for further details.

PR #1724, opened 2019-07-11
Adds support for fragments in userAuthorizationUri

DeCaMil:
Fixes [#959](https://github.com/spring-projects/spring-security-oauth/issues/959)
Adds support for fragments in `userAuthorizationUri` in `OAuth2ClientContextFilter `


jgrandja:
Thanks for the PR @DeCaMil.

As per spec, in section [3.1.2. Redirection Endpoint](https://tools.ietf.org/html/rfc6749#section-3.1.2):

> The redirection endpoint URI MUST be an absolute URI as defined by
   [RFC3986] Section 4.3.  The endpoint URI MAY include an
   "application/x-www-form-urlencoded" formatted (per Appendix B) query
   component ([RFC3986] Section 3.4), which MUST be retained when adding
   additional query parameters.  **The endpoint URI MUST NOT include a
   fragment component**.

As indicated in bold, a fragment component is not allowed. I'm going to close this PR based on this.

Thank you for your understanding.

Issue #1759, opened 2019-10-14
Unsafe deserialization in JDBC services


artem-smotrakov:
I just realised that the following classes have a similar problem which is described in #1682 

- JdbcTokenStore
- JdbcClientTokenServices
- JdbcAuthorizationCodeServices

Making a long story short, compromising a SQL database server may lead to compromising the whole application. It can happen due to unsafe deserialization in `SerializationUtils` class.

Of course, compromising a database looks like a serious issue by itself. Database servers should be protected and configured in a secure way. But nevertheless I am wondering if spring-security-oauth should be updated a bit to prevent arbitrary code execution in the application in case databases are hacked.

I'll prepare a patch similar to #1703

vaddemgen:
@jgrandja @artem-smotrakov 

It doesn't work for me. When I tried to serialize my own ```org.springframework.security.core.userdetails.UserDetails```, I get the following problem:

```java
java.lang.IllegalArgumentException: java.io.NotSerializableException: Not allowed to deserialize my.own.UserDetailsImpl
```

It happens because my custom implementation is NOT in the list [`ALLOWED_CLASSES`](https://github.com/spring-projects/spring-security-oauth/commit/b214a25fdd8eae34c8ad803e7022fd17d1c63eb6#diff-28d3f6d7939c6931299a5d3c5e632d58R26). This list is immutable (final and unmodifiable) so I can't add my own class there.

How to add my own implementation to the `ALLOWED_CLASSES` list so that it is allowed?

artem-smotrakov:
Thanks for catching this @vaddemgen 

I see two ways to solve it.

1. Introduce a way how the allowed class list can be updated by an application.
2. In `SaferObjectInputStream`, inspect all classes and interfaces which the class inherits, and allow deserialization if `my.own.UserDetailsImpl` inherits or implements one of the allowed classes.

The second way may significally extend the number of allowed classes. I would prefer the first way. However, `2.3.7` is going to be a patch release which should not add any public API (see [this comment](https://github.com/spring-projects/spring-security-oauth/pull/1760#issuecomment-541927790) from @jgrandja).

I can open a pull request after we decide how to address the problem.

jgrandja:
Thanks @vaddemgen.

@artem-smotrakov I agree, the 2nd option is not ideal so I think we should avoid it. Another option is introduce a discoverable well-known file in the classpath that can be configured by the user for the whitelist.

[...]

PR #1760, opened 2019-10-14
Safer deserialization in JDBC services

artem-smotrakov:
Please review the following patch which fixes gh-1759 (similar to #1703)

- Updated `SerializationUtils` to use `SaferObjectInputStream` with a list of allowed classes.
- Added `SerializationUtilsTest`.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationUtils.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationUtils.java
@@ -3,12 +3,31 @@ package org.springframework.security.oauth2.common.util;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.NotSerializableException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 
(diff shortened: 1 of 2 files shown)


artem-smotrakov:
@jgrandja I just realized that several JDBC classes are also affected by the issue that was recently fixed in `RedisTokenStore`. Would it be possible to integrate this patch to the next release?

I moved `SaferObjectInputStream` to another package, and made it public, since it's now used by two classes. Please let me know if the javadoc needs to be improved.

jgrandja:
@artem-smotrakov The reason I made `SaferObjectInputStream` package-private is that we cannot expose a public API in a [patch release](https://apr.apache.org/versioning.html#strategy).

We're not looking to release any further minor releases, just patches that contain bug fixes and minor enhancements. As you know, this project is in maintenance mode in favour of the new OAuth support in Spring Security 5.x.

If you really need this to go into the next release I would be ok merging it but it cannot expose any public API's. This is one of those instances where duplicating `SaferObjectInputStream` either as package-private or private inner class would be fine.

artem-smotrakov:
@jgrandja No problem, I'll update the patch to hide `SaferObjectInputStream`.

artem-smotrakov:
@jgrandja `SerializationUtils` now has its own `SaferObjectInputStream`. Please let me know if I missed something.

jgrandja:
Thanks again for the PR @artem-smotrakov. This is now in master.

PR #1770, opened 2019-10-15
Dependency upgrades

jzheaux:


samples/oauth2/sparklr/pom.xml
```diff
@@ -15,15 +15,15 @@
 
 	<properties>
 		<m2eclipse.wtp.contextRoot>/sparklr2</m2eclipse.wtp.contextRoot>
-		<jackson2.version>2.9.8</jackson2.version>
+		<jackson2.version>2.9.10</jackson2.version>
```
jgrandja: @jzheaux Let's update to the latest which is 2.10.0

spring-security-oauth2/pom.xml
```diff
@@ -16,7 +16,7 @@
 		<jackson1.version>1.9.13</jackson1.version>
 		<jackson2.version>2.9.10</jackson2.version>
 		<servlet-api.version>3.0.1</servlet-api.version>
-		<spring.security.jwt.version>1.0.9.RELEASE</spring.security.jwt.version>
+		<spring.security.jwt.version>1.0.10.RELEASE</spring.security.jwt.version>
```
jgrandja: @jzheaux This will need to be updated to 1.0.11. NOTE: 1.0.10 was released Jan 24, 2019.

jgrandja:
This has already been merged

PR #1784, opened 2019-10-21
Allow specifying allowed classes for deserialization

### LINKED ISSUE #1783: Make deserialization great again
The updates #1759 and #1682 for deserialization procedures don't allow working with custom implementations. See https://github.com/spring-projects/spring-security-oauth/issues/1759#issuecomment-543114291 for details.

Applications should be able to customize a list of allowed classes for deserialization.

artem-smotrakov:
This patch fixes #1783

1. Moved `SaferObjectInputStream` to `org.springframework.security.oauth2.common.util`. The class allow specifying custom list of allowed classes.
2. Added `SerializationStrategy` based on `SerializationUtils`. The new class uses `SaferObjectInputStream` and defines the default list of allowed classes for deserialization. The class allow specifying custom list of allowed classes.
3. Revert changes in `SerializationUtils` and marked it as deprecated.
4. Updated JDBC services to use `SerializationStrategy` instead of `SerializationUtils`. The classes allow specifying custom `SerializationStrategy`.
5. Updated `JdkSerializationStrategy` (used with Redis)` to use `SerializationStrategy`. The class allow specifying custom list of allowed classes.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationStrategy.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationStrategy.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.common.util;
+
+/**
+ * Defines how objects are serialized an deserialized.
+ *
+ * @author Artem Smotrakov
+ * @since 2.4
+ */
+public interface SerializationStrategy {
+
+    /**
+     * Serializes an object.
+     *
+     * @param object The object to be serialized.
+     * @return A byte array.
+     */
+    byte[] serialize(Object object);
+
+    /**
+     * Deserializes an object from a byte array.
+     *
+     * @param byteArray The byte array.
+     * @param <T>       The type of the object.
+     * @return The deserialized object.
+     */
+    <T> T deserialize(byte[] byteArray);
+
+}
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationUtils.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationUtils.java
@@ -1,64 +1,59 @@
 package org.springframework.security.oauth2.common.util;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
+import org.springframework.core.io.support.SpringFactoriesLoader;
+import org.springframework.util.Assert;
 
-import org.springframework.core.ConfigurableObjectInputStream;
+import java.util.List;
 
+/**
+ * This is a helper class for serializing and deserializing objects with a {@link SerializationStrategy}.
+ * The class looks for the strategy in {@code META-INF/spring.factories},
+ * or the strategy can be also set by calling {@link #setSerializationStrategy(SerializationStrategy)} method.<br/>
+ * If no strategy specified, the class uses {@link DefaultSerializationStrategy}.
+ * <br/>
+ * Note that the default strategy allows deserializing arbitrary classes which may result to security problems
+ * if data comes from an untrusted source. To prevent possible issues, use {@link WhitelistedSerializationStrategy}
+ * with a list of allowed classes for deserialization.
+ */
 public class SerializationUtils {
 
-	public static byte[] serialize(Object state) {
-		ObjectOutputStream oos = null;
-		try {
-			ByteArrayOutputStream bos = new ByteArrayOutputStream(512);
-			oos = new ObjectOutputStream(bos);
-			oos.writeObject(state);
-			oos.flush();
-			return bos.toByteArray();
-		}
-		catch (IOException e) {
-			throw new IllegalArgumentException(e);
-		}
-		finally {
-			if (oos != null) {
-				try {
-					oos.close();
-				}
-				catch (IOException e) {
-					// eat it
-				}
-			}
-		}
-	}
+    private static SerializationStrategy strategy = new DefaultSerializationStrategy();
 
-	public static <T> T deserialize(byte[] byteArray) {
-		ObjectInputStream oip = null;
-		try {
-			oip = new ConfigurableObjectInputStream(new ByteArrayInputStream(byteArray),
-					Thread.currentThread().getContextClassLoader());
-			@SuppressWarnings("unchecked")
-			T result = (T) oip.readObject();
-			return result;
-		}
-		catch (IOException e) {
-			throw new IllegalArgumentException(e);
-		}
-		catch (ClassNotFoundException e) {
-			throw new IllegalArgumentException(e);
-		}
-		finally {
-			if (oip != null) {
-				try {
-					oip.close();
-				}
-				catch (IOException e) {
-					// eat it
-				}
-			}
-		}
-	}
+    static {
+        List<SerializationStrategy> strategies = SpringFactoriesLoader.loadFactories(
+                SerializationStrategy.class, SerializationUtils.class.getClassLoader());
+        if (strategies.size() > 1) {
+            throw new IllegalArgumentException(
+                    "Too many serialization strategies in META-INF/spring.factories");
+        }
+        if (strategies.size() == 1) {
+            strategy = strategies.get(0);
+        }
+    }
+
+    /**
+     * @return The current serialization strategy.
+     */
+    public static SerializationStrategy getSerializationStrategy() {
+        return strategy;
+    }
+
+    /**
+     * Sets a new serialization strategy.
+     *
+     * @param serializationStrategy The serialization strategy.
+     */
+    public static void setSerializationStrategy(SerializationStrategy serializationStrategy) {
+        Assert.notNull(serializationStrategy, "Serialization strategy can't be null");
+        strategy = serializationStrategy;
+    }
+
+    public static byte[] serialize(Object object) {
+        return strategy.serialize(object);
+    }
+
+    public static <T> T deserialize(byte[] byteArray) {
+        return strategy.deserialize(byteArray);
+    }
 
 }
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/WhitelistedSerializationStrategy.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/WhitelistedSerializationStrategy.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.common.util;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.NotSerializableException;
+import java.io.ObjectInputStream;
+import java.io.ObjectStreamClass;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.springframework.util.ClassUtils;
+
+/**
+ * This is a {@link SerializationStrategy} which uses a whitelist of allowed classes for deserialization.
+ *
+ * @author Artem Smotrakov
+ * @since 2.4
+ */
+public class WhitelistedSerializationStrategy extends DefaultSerializationStrategy {
+
+    /**
+     * A list of classes which are allowed to deserialize by default.
+     */
+    private static final List<String> DEFAULT_ALLOWED_CLASSES;
+
+    static {
+        List<String> classes = new ArrayList<String>();
+        classes.add("java.lang.");
+        classes.add("java.util.");
+        classes.add("org.springframework.security.");
+        DEFAULT_ALLOWED_CLASSES = Collections.unmodifiableList(classes);
+    }
+
+    /**
+     * A list of classes which allowed to deserialize.
+     */
+    private final List<String> allowedClasses;
+
+    /**
+     * Initializes {@link WhitelistedSerializationStrategy} with the list of classes
+     * which are allowed to deserialize by default.
+     */
+    public WhitelistedSerializationStrategy() {
+        this(DEFAULT_ALLOWED_CLASSES);
+    }
+
+    /**
+     * Initializes {@link WhitelistedSerializationStrategy} with specified allowed classes.
+     *
+     * @param allowedClasses The allowed classes for deserialization.
+     */
+    public WhitelistedSerializationStrategy(List<String> allowedClasses) {
+        this.allowedClasses = Collections.unmodifiableList(allowedClasses);
+    }
+
+    protected ObjectInputStream createObjectInputStream(byte[] byteArray) throws IOException {
+        return new WhitelistedObjectInputStream(new ByteArrayInputStream(byteArray),
+                Thread.currentThread().getContextClassLoader(), allowedClasses);
+    }
+
+    /**
+     * Special ObjectInputStream subclass that checks if classes are allowed to deserialize. The class
+     * should be configured with a whitelist of only allowed (safe) classes to deserialize.
+     */
+    private static class WhitelistedObjectInputStream extends ObjectInputStream {
+
+        /**
+         * The list of classes which are allowed for deserialization.
+         */
+        private final List<String> allowedClasses;
+
+        /**
+         * The class loader to use for loading local classes.
+         */
+        private final ClassLoader classLoader;
+
+        /**
+         * Create a new WhitelistedObjectInputStream for the given InputStream, class loader and
+         * allowed class names.
+         *
+         * @param in             The InputStream to read from.
+         * @param classLoader    The ClassLoader to use for loading local classes.
+         * @param allowedClasses The list of allowed classes for deserialization.
+         * @throws IOException If something went wrong.
+         */
+        private WhitelistedObjectInputStream(InputStream in, ClassLoader classLoader, List<String> allowedClasses)
+                throws IOException {
+            super(in);
+            this.classLoader = classLoader;
+            this.allowedClasses = Collections.unmodifiableList(allowedClasses);
+        }
+
+        /**
+         * Resolve the class only if it's allowed to deserialize.
+         *
+         * @see ObjectInputStream#resolveClass(ObjectStreamClass)
+         */
+        @Override
+        protected Class<?> resolveClass(ObjectStreamClass classDesc)
+                throws IOException, ClassNotFoundException {
+            if (isProhibited(classDesc.getName())) {
+                throw new NotSerializableException("Not allowed to deserialize " + classDesc.getName());
+            }
+            if (this.classLoader != null) {
+                return ClassUtils.forName(classDesc.getName(), this.classLoader);
+            }
+            return super.resolveClass(classDesc);
+        }
+
+        /**
+         * Check if the class is allowed to be deserialized.
+         *
+         * @param className The class to check.
+         * @return True if the class is not allowed to be deserialized, false otherwise.
+         */
+        private boolean isProhibited(String className) {
+            for (String allowedClass : this.allowedClasses) {
+                if (className.startsWith(allowedClass)) {
+                    return false;
+                }
+            }
+            return true;
+        }
+    }
+}
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/JdkSerializationStrategy.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/JdkSerializationStrategy.java
@@ -1,26 +1,50 @@
 package org.springframework.security.oauth2.provider.token.store.redis;
 
-import org.springframework.data.redis.serializer.JdkSerializationRedisSerializer;
+import org.springframework.core.serializer.support.SerializationFailedException;
+
+import java.io.Serializable;
+
+import org.springframework.security.oauth2.common.util.SerializationUtils;
 
 /**
- * Serializes objects using {@link JdkSerializationRedisSerializer}
+ * Serializes and deserializes allowed objects using {@link SerializationUtils}.
  *
  * @author efenderbosch
- *
+ * @author Artem Smotrakov
  */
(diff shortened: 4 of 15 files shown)


artem-smotrakov:
@vaddemgen Could you please take a look at the patch?

vaddemgen:
@artem-smotrakov **I have just checked and it works. **

**_But I have a question about using this deserialization mechanism_**.

Could you check if packages are properly added to the list of allowed during deserialization? I think, that was the point.

I had to add changes to the code. Tthe other implementation of ```TokenStore``` (`JdbcTokenStore`) was used. Here, during the initialization of the TokenStore bean, an instance of SerializationStrategy was created with my list of allowed classes:

```java
package my.project.config.oauth2;

import my.project.model.CustomPersonDetails;
import org.springframework.security.oauth2.common.util.SerializationStrategy;
import org.springframework.security.oauth2.provider.token.store.JdbcTokenStore;
// ...

@Configuration
@EnableAuthorizationServer
public class AuthorizationServerConfiguration extends AuthorizationServerConfigurerAdapter {

  // ...

  @Bean
  public TokenStore tokenStore() {
    // I use JdbcTokenStore instead of the default InMemoryTokenStore
    JdbcTokenStore jdbcTokenStore = new JdbcTokenStore(dataSource);
    jdbcTokenStore.setSerializationStrategy(new SerializationStrategy(
        List.of(
            // Here I add the packages provided by the constant 
            // SerializationStrategy#DEFAULT_ALLOWED_CLASSES
            "java.lang.",
            "java.util.",
            "org.springframework.security.",
            // And I add my own implementation of PersonDetails 
            // to the list of allowed packages during the deserialization
            CustomPersonDetails.class.getPackageName()
        )
    ));
    return jdbcTokenStore;
  }

  // ...
}
```

It seems to me it's better to also add the ability to customize this list by an env variable.

artem-smotrakov:
@vaddemgen Thanks for the feedback and testing the patch.

> Could you check please if packages are properly added to the list of allowed during deserialization? 

Your code looks correct to me. Do you mean the default list of allowed classes should be extended?

> It seems to me it's better to also add the ability to customize this list by an env variable.

I think it may be a nice improvement. Alternatively, we can introduce a system property to customize the list.

@jgrandja What do you think?

artem-smotrakov:
If we have a system property or an environment variable, then application code may not be updated which is good.

vaddemgen:
@artem-smotrakov 

> Do you mean the default list of allowed classes should be extended?

No, It looks good. Thank you for the answer!

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SaferObjectInputStream.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.List;
+import org.springframework.util.ClassUtils;
+
+/**
+ * Special ObjectInputStream subclass that checks if classes are allowed to deserialize.
+ * The class should be configured with a whitelist of only allowed (safe) classes to deserialize.
+ *
+ * @author Artem Smotrakov
+ */
+public class SaferObjectInputStream extends ObjectInputStream {
```
jgrandja: Let's rename this to `WhitelistedObjectInputStream`.

Can we also make this a `private class` internal (inner) to `SerializationStrategy`.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationStrategy.java
```diff
@@ -0,0 +1,102 @@
...
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Serialization strategy based on {@link SaferObjectInputStream}.
+ */
+public class SerializationStrategy {
```
jgrandja: Let's make this an `interface` and name the whitelisted implementation `WhitelistedSerializationStrategy` or `DefaultSerializationStrategy`

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationUtils.java
```diff
@@ -8,6 +8,10 @@
 
 import org.springframework.core.ConfigurableObjectInputStream;
 
+/**
+ * This class should not be used since it uses unsafe deserialization via {@link ConfigurableObjectInputStream}.
+ */
+@Deprecated
```
jgrandja: `SerializationUtils` is still used in certain places. We should ensure there is no further usages since `SerializationStrategy` is now preferred.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/JdkSerializationStrategy.java
```diff
@@ -1,26 +1,89 @@
...
+  public JdkSerializationStrategy() {
+    this.serializationStrategy = new SerializationStrategy();
+  }
+
+  /**
+   * Initializes a {@link JdkSerializationStrategy} with a list of allowed classes.
+   *
+   * @param allowedClasses The allowed classes.
+   */
+  public JdkSerializationStrategy(List<String> allowedClasses) {
```
jgrandja: Please remove this constructor and add `setSerializationStrategy(SerializationStrategy serializationStrategy)`, now that `SerializationStrategy`  is an interface

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationUtils.java
artem-smotrakov: Right, I missed several places where it's used. I'll replace `SerializationUtils` with `WhitelistedSerializationStrategy`.

I also noticed that some tests use `org.springframework.util.SerializationUtils`. I guess it can be also replaced with `WhitelistedSerializationStrategy`.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationStrategy.java
artem-smotrakov: Okay, no problem.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/JdkSerializationStrategy.java
artem-smotrakov: Okay, I can do that. I thought it might be better to make `serializationStrategy` filed final.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SaferObjectInputStream.java
artem-smotrakov: Sure, no problem. However, it can't be a private class since `SerializationStrategy` is now an interface.

artem-smotrakov: I'll restrict the access to package-private.

artem-smotrakov:
Thanks for the review @jgrandja I have addressed the comments, please take a look.

We can also introduce a way how a user can set a list of allowed classes via environment variable or system property (please see the comment above from @vaddemgen). What do you think?

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationStrategy.java
```diff
@@ -0,0 +1,99 @@
...
+   * @param <T> The type of the object.
+   * @return The deserialized object.
+   */
+  <T> T deserialize(byte[] byteArray);
+
+  /**
+   * Special ObjectInputStream subclass that checks if classes are allowed to deserialize.
+   * The class should be configured with a whitelist of only allowed (safe) classes to deserialize.
+   */
+  class WhitelistedObjectInputStream extends ObjectInputStream {
```
jgrandja: Let's move this to `WhitelistedSerializationStrategy` as a `private class` since it's part of that implementation.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationStrategy.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.ObjectInputStream;
+import java.io.ObjectStreamClass;
+import java.util.Collections;
+import java.util.List;
+import org.springframework.util.ClassUtils;
+
+/**
+ * Defines how objects are serialized an deserialized.
+ *
+ * @author Artem Smotrakov
```
jgrandja: Add `@since 2.4`

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationUtils.java
```diff
@@ -8,6 +8,10 @@
 
 import org.springframework.core.ConfigurableObjectInputStream;
 
+@Deprecated
```
jgrandja: I'm debating whether we should deprecate. I feel there are a lot of touch points with `SerializationStrategy` via `setSerializationStrategy()` in the various collaborators. My preference would be to touch less of the code. Since all collaborators use `SerializationUtils` it might be better to isolate the changes there.

We could add a `SerializationStrategy` static member and default to `WhitelistedSerializationStrategy`. The application could override the behaviour via these options:

1. Add new `public static void setSerializationStrategy(SerializationStrategy serializationStrategy)`
2. Using `spring.factories`

With `spring.factories`, the user could override by adding a classpath resource under `META-INF/spring.factories` with the contents being:

`org.springframework.security.oauth2.common.util.SerializationStrategy = mypackage.CustomSerializationStrategy`

You can use `SpringFactoriesLoader` to check if the user has supplied a custom `SerializationStrategy`.

What are your thoughts on this?

jgrandja:
@artem-smotrakov 

> We can also introduce a way how a user can set a list of allowed classes via environment variable or system property...

This can complicate things since we would need to specify a specific format so it can be read correctly from the environment. I'd rather keep things simple. Based on my previous comment on proposing to add `SerializationUtils.setSerializationStrategy()`, this will provide the ability for the application to read the allowed classes from the environment or system property or any source and than configure it with the proposed setter.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationUtils.java
artem-smotrakov: Okay, let me try to add a `SerializationStrategy` static member to `SerializationUtils` and a new  `SerializationUtils.setSerializationStrategy()` method. Looks like it can make simplify the patch which is definitely good.

I am planning to get back to it on Nov 4th. I think we have time to finish it the next week. Thanks for the comments!

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationStrategy.java
artem-smotrakov: Okay, no problem.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationStrategy.java
artem-smotrakov: Sure.

byference:
I found a problem in `JdkSerializationStrategy#ALLOWED_CLASSES` ，when I customize `UserDetails ` implementation (it is not in  `ALLOWED_CLASSES`  list) and use `RedisTokenStore` on `AuthorizationServerConfiguration#configure(AuthorizationServerEndpointsConfigurer)`，will throw `NotSerializableException`

~~~java
// org.springframework.security.oauth2.provider.token.store.redis.SaferObjectInputStream#resolveClass
protected Class<?> resolveClass(ObjectStreamClass classDesc) throws IOException, ClassNotFoundException {
    if (isProhibited(classDesc.getName())) {
        throw new NotSerializableException("Not allowed to deserialize " + classDesc.getName());
    }
    return super.resolveClass(classDesc);
}

private boolean isProhibited(String className) {
    for (String allowedClass : this.allowedClasses) {
        if (className.startsWith(allowedClass)) {
            return false;
        }
    }
    return true;
}
~~~


my `spring-security-oauth2 ` is ` 2.3.7.RELEASE` .

artem-smotrakov:
@jgrandja I have addressed your comments. Please take a look.

artem-smotrakov:
@vaddemgen @byference 

You can now implement a custom `SerializationStrategy` which allows deserializing classes which you need. For example, you can extend `WhitelistedSerializationStrategy`:

```java
public class CustomSerializationStrategy extends WhitelistedSerializationStrategy {

    private static final List<String> ALLOWED_CLASSES = new ArrayList<>();
    static {
        ALLOWED_CLASSES.add("java.lang.");
        ALLOWED_CLASSES.add("java.util.");
        ALLOWED_CLASSES.add("org.springframework.security.");
        ALLOWED_CLASSES.add("your.custom.impl.");
    }

    public CustomSerializationStrategy() {
        super(ALLOWED_CLASSES);
    }
}
```

Then, you can set your serialization strategy with `SerializationUtils.setSerializationStrategy()` method. Or, you can define it in the `META-INF/spring.factories` file:

```
org.springframework.security.oauth2.common.util.SerializationStrategy = custom.impl.CustomSerializationStrategy
```

I tested it with a demo application, it seems to work. Could you please try it out as well?

byference:
`SerializationUtils.setSerializationStrategy()`  and `org.springframework.security.oauth2.common.util.SerializationStrategy` is not available in the latest version (`2.3.7.RELEASE `) . 

now, I can customizing the implementation of `RedisTokenStoreSerializationStrategy` （refer to`JdkSerializationStrategy`）to solve my problem,

~~~java
public class CustomJdkSerializationStrategy extends StandardStringSerializationStrategy {
    ...
    static {
        List<String> classes = new ArrayList<String>();
        classes.add("java.lang.");
        classes.add("java.util.");
        classes.add("org.springframework.security.");
        classes.add("my.custom.impl.");
        ALLOWED_CLASSES = Collections.unmodifiableList(classes);
    }

    @Override
    @SuppressWarnings("unchecked")
    protected <T> T deserializeInternal(byte[] bytes, Class<T> clazz) {
        ...
        try {
            SaferObjectInputStream saferObjectInputStream = new SaferObjectInputStream(new ByteArrayInputStream(bytes), ALLOWED_CLASSES);
            return (T) saferObjectInputStream.readObject();
        } catch (Exception e) {
            throw new SerializationFailedException("Failed to deserialize payload", e);
        }
    }
    ...
    // copied from org.springframework.security.oauth2.provider.token.store.redis.SaferObjectInputStream
    class SaferObjectInputStream extends ObjectInputStream {
        ...
    }
}


@Configuration
@EnableAuthorizationServer
protected static class AuthorizationServerConfiguration extends AuthorizationServerConfigurerAdapter {
    ...
    @Override
    public void configure(AuthorizationServerEndpointsConfigurer endpoints) throws Exception {
        RedisTokenStore redisTokenStore = new RedisTokenStore(redisConnectionFactory);
        redisTokenStore.setSerializationStrategy(new CustomJdkSerializationStrategy());
        endpoints.tokenStore(redisTokenStore)
            .authenticationManager(authenticationManager);
    }
}
~~~

but i don't think it's a good way, so looking forward to the new version @artem-smotrakov

artem-smotrakov:
@byference The issue you are facing is #1783. But this is a pull request which addresses this issue. The pull request is targeted to `2.4.0`.

artem-smotrakov:
@jgrandja 

> Also, can you add a test for `JdbcClientTokenServices`

I added a couple of test cases for `JdbcClientTokenServicesTests`.

spring-security-oauth2/src/test/java/org/springframework/security/oauth2/common/util/SerializationUtilsTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.UUID;
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Artem Smotrakov
+ */
+public class SerializationUtilsTest {
```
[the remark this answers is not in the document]
artem-smotrakov: @jgrandja I added `CustomSerializationStrategyTest` which mocks `SpringFactoriesLoader.loadFactories()`. I borrowed the idea from 

https://github.com/spring-projects/spring-security/blob/00f8991facc6b615d6e2824c9b9ceff7b415d57e/config/src/test/java/org/springframework/security/config/annotation/web/WebSecurityConfigurerAdapterPowermockTests.java

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/SerializationUtils.java
```diff
@@ -1,64 +1,73 @@
...
+        }
+        if (strategies.size() == 1) {
+            strategy = strategies.get(0);
+        }
+    }
+
+    /**
+     * @return The current serialization strategy.
+     */
+    public static SerializationStrategy getSerializationStrategy() {
```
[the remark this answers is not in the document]
jgrandja: Ok we can leave as-is.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/WhitelistedSerializationStrategy.java
```diff
@@ -0,0 +1,146 @@
...
+
+import org.springframework.util.ClassUtils;
+
+/**
+ * This is a {@link SerializationStrategy} which uses a whitelist of allowed classes for deserialization.
+ *
+ * @author Artem Smotrakov
+ * @since 2.4
+ */
+public class WhitelistedSerializationStrategy implements SerializationStrategy {
```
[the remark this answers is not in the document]
jgrandja: > a user can extend this class to use it

Even if a user were to extend this class, there aren't any `protected` methods where a user can override the behaviour. They would only be able to override `serialize` and/or `deserialize` to change behaviour. Instead of extension, a user can implement a delegation-based approach to achieve the same.

> Do you have a reason why this class has to be final?

To prevent extension so we are not on the hook for support. However, given that 2.4.0 is the last minor version and this project will soon be deprecated, it's pretty much a moot point. I would be fine if we left this as-is.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/WhitelistedSerializationStrategy.java
```diff
@@ -0,0 +1,162 @@
...
+
+        /**
+         * Resolve the class only if it's allowed to deserialize.
+         *
+         * @see ObjectInputStream#resolveClass(ObjectStreamClass)
+         */
+        @Override
+        protected Class<?> resolveClass(ObjectStreamClass classDesc)
+                throws IOException, ClassNotFoundException {
+            if (isProhibited(classDesc.getName())) {
```
jgrandja: As it stands now, when an application upgrades to 2.4 and have extended `OAuth2AccessToken` (or `OAuth2RefreshToken` and `OAuth2Authentication`), it will be forced to configure a custom `WhitelistedSerializationStrategy`, otherwise the application will fail when attempting to deserialize.

The preference would be that the application will continue to work after upgrading to 2.4 without having to provide an additional configuration (code change). Can we add an additional check for superclass? For example, `classDesc.forClass().getSuperclass()`

artem-smotrakov:
> Also, I would like to make you aware (so there are no surprises), that the main reason for releasing 2.4 is to officially `@Deprecated` all classes in `spring-security-oauth2`. This will also include the new classes introduced in this PR. However, the 2.4.x branch will be supported until approx. March 2021 so this improvement is still valuable until than. Keep an eye on the blog post next week (same time as release Nov 14) for further details.

Sure, thanks for letting me know!

spring-security-oauth2/src/test/java/org/springframework/security/oauth2/client/token/JdbcClientTokenServicesTests.java
```diff
@@ -79,4 +85,37 @@ public void testSaveAndRemoveToken() throws Exception {
...
+		List<String> allowedClasses = new ArrayList<String>();
+		allowedClasses.add("java.util.");
+		allowedClasses.add("org.springframework.security.");
+		allowedClasses.add("org.company.oauth2.CustomOAuth2AccessToken");
+		allowedClasses.add("org.company.oauth2.CustomOAuth2Authentication");
+		WhitelistedSerializationStrategy newStrategy = new WhitelistedSerializationStrategy(allowedClasses);
+		SerializationStrategy oldStrategy = SerializationUtils.getSerializationStrategy();
+		try {
+			SerializationUtils.setSerializationStrategy(newStrategy);
+			OAuth2AccessToken accessToken = new DefaultOAuth2AccessToken("FOO");
```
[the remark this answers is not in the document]
artem-smotrakov: Yes, good catch! I also forget to add a couple of checks in this test case.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/WhitelistedSerializationStrategy.java
artem-smotrakov: @jgrandja 

> As it stands now, when an application upgrades to 2.4 and have extended `OAuth2AccessToken` (or `OAuth2RefreshToken`, `OAuth2Authentication`), it will be forced to configure a custom `WhitelistedSerializationStrategy`, otherwise the application will fail when attempting to deserialize.

Unfortunately yes, that's how it's going to be.

> The preference would be that the application will continue to work after upgrading to 2.4 without having to provide an additional configuration (code change). 

Yes, ideally.

> Can we add an additional check for superclass? For example, `classDesc.forClass().getSuperclass()`

I am not sure if we should do that. If we add such a check, an attacker can prepare a class `MaliciousOAuth2AccessToken` which extends `DefaultOAuth2AccessToken`, and put a malicious code to the constructor. Then, `MaliciousOAuth2AccessToken` can pass the `is Prohibited()` check.

jgrandja: @artem-smotrakov 

> I am not sure if we should do that. If we add such a check, an attacker can prepare a class `MaliciousOAuth2AccessToken`...

Yes, you're right. Ok, so we should not apply check on superclass?

> As it stands now, when an application upgrades to 2.4 and have extended `OAuth2AccessToken` (or `OAuth2RefreshToken`, `OAuth2Authentication`), it will be forced to configure a custom `WhitelistedSerializationStrategy`, otherwise the application will fail when attempting to deserialize.

This change introduces incompatibility changes between 2.3.x and 2.4.0, which cannot happen in a minor  release. An application built against 2.3.x should be able to upgrade to 2.4.0 without any issues. This change would be fine for a major version though. Please see [Versioning](https://apr.apache.org/versioning.html#strategy) for additional details.

In order to get this change into 2.4.0, this will need to be an "opt-in" feature rather than the default configuration. So if the application wants stricter deserialization, than they would need to explicitly configure `WhitelistedSerializationStrategy`. For the default configuration, the application should perform the same as it did in 2.3.x.

artem-smotrakov: @jgrandja 

> In order to get this change into 2.4.0, this will need to be an "opt-in" feature rather than the default configuration.

That's sad. I would prefer to make it a bit more secure by default. I'll update it to be an opt-it feature, however I don't like it.

jgrandja: @artem-smotrakov 

> I would prefer to make it a bit more secure by default

I definitely would too. But with the way versioning works, this would break existing applications and users would log new issues around this as we have seen with the 2.3.7 release.

artem-smotrakov: @jgrandja 

* I added a new `DefaultSerializationStrategy` which allows all classes by default. The `SerializationUtils` uses `DefaultSerializationStrategy` by default.
* I also mentioned in javadoc that the default settings may be unsafe if data come from an untrusted source.
* Updated tests

Please let me know what you think.

jgrandja:
Thanks you @artem-smotrakov ! This is now in master.

artem-smotrakov:
> Thank you @artem-smotrakov ! This is now in master.

Awesome, thanks for the review @jgrandja !

Issue #996, opened 2017-03-01
Remove jackson-mapper-asl from spring-security-oauth2


rdifrango:
Any reason that `jackson-mapper-asl` can't be removed from `spring-security-oauth2`, see:

```
<dependency>
	<groupId>org.codehaus.jackson</groupId>
	<artifactId>jackson-mapper-asl</artifactId>
	<version>${jackson1.version}</version>
</dependency>
```

romeara:
As a note, the Maven repository page details that this component is now jackson-databind: https://mvnrepository.com/artifact/org.codehaus.jackson/jackson-mapper-asl. 

Even if it can't be removed, updating this to the new group/artifact ID would allow Maven/Gradle dependency resolution to handle conflicting version dependencies in a more consistent way

Kitanotori:
Also wondering if there's still need for such an old version of Jackson?

greyfairer:
This issue bit me today. If you use the Jackson 2 combined with spring-security-oauth2, the org.codehaus... annotations are ignored, only the com.fasterxml... work.

Easy mistake to make if your IDE is doing imports automatically.

wilx:
Can this be resolved by removing the old dependency in favour of the new one?

elion:
Any news on this?

OrangeDog:
@joshlandin are you sure? All indications are that they're all for jackson-databind, not jackson-mapper-asl.

jgrandja:
@joshlandin We'll be removing Jackson 1 in the 2.4.0 release, which is scheduled for Nov 7.

davydotcom:
can we remove this from 2.0.x as this library has critical CVE vulnerabilities in the first place

jgrandja:
@davydotcom 2.0.x is no longer supported. The currently supported branches are 2.3.x and 2.4.x (2.4.0 releases this Thur Nov 14).

davydotcom:
Considering its a critical CVE I think it should be handled. We are in the midst of updating to latest spring framework as well. Also the PR is a clean cherry-pick and does work

OrangeDog:
@davydotcom which CSV? As I said above, I believe there are a bunch incorrectly marked for jackson-mapper-asl that are actually only in jackson-databind. Would need confirmation from jackson developers to be completely sure.

PeterWippermann:
[...]

PR #1791, opened 2019-10-25
Remove Jackson 1 Support

jzheaux:
Fixes gh-996


jgrandja:
Thanks for the PR @jzheaux. This is now in master.

Issue #1836, opened 2020-03-14
RedisTokenStore should delete access_to_refresh key when removing refresh token


mashice:
https://github.com/spring-projects/spring-security-oauth/blob/d72922e395aa9c2fd0fd5b2de2e557790c2b92c5/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/redis/RedisTokenStore.java#L364

in line 364 want to delete refreshToken,  redis key  is ACCESS_TO_REFRESH + refreshToken, but in line 213 save key as ACCESS_TO_REFRESH + accessToken.

![image](https://user-images.githubusercontent.com/40679671/76675246-46a60d80-65f2-11ea-83df-ffee6c16643c.png)
![image](https://user-images.githubusercontent.com/40679671/76675237-2fffb680-65f2-11ea-89a3-bcaf2c0d0e91.png)

srempfer:
The storing part is correct but the access2RefreshKey is calculated wrong for removing part.

Scenario 1: => no key is left behind
1. removeAccessToken
2. removeRefreshToken

Scenario 2: => no key is left behind
1. removeRefreshToken
2. removeAccessToken

Scenario 3: => key for access_to_refresh is left behind
1. removeRefreshToken

I develop currently a test and fix for that. PR will follow...

PR #1848, opened 2020-04-16
Delete access_to_refresh key when removing refresh token in RedisTokenStore

srempfer:
fixes gh-1836


srempfer:
Is there a general problem with the build:

https://travis-ci.org/github/spring-projects/spring-security-oauth/jobs/675911859#L2523-L2539

Or did I missed something?

jgrandja:
Thanks for the PR @srempfer. This is now in master and backported to 2.4.x.

PR #1856, opened 2020-05-01
Added a custom jwtconverter compatible to jdbc store.

mplescano:
I have implemented an oauth server using the jdbc storage along side a custom jwt converter.
I have fixed some integration test that failed randomly.


jgrandja:
@mplescano The `JwtJdbcTokenConverter` in this PR does not address a specific issue with `JwtAccessTokenConverter`. I understand your preference is too eliminate the dependency the refresh token has with the access token as per [comment](https://github.com/spring-projects/spring-security-oauth/pull/1573#issuecomment-622075352). The fact is that there is no issue with `JwtAccessTokenConverter` so I won't be able to merge this PR.

As I mentioned in the comment:

> It doesn't stop you from applying this change on your end. All you need to do is configure a custom `JwtAccessTokenConverter` (or `TokenEnhancer`).

You can simply configure `JwtJdbcTokenConverter` in your application and it will deliver the format you are requiring.

Issue #1857, opened 2020-05-04
DefaultTokenServices should generate url-safe tokens


torstenkuhnhenne:
### Summary
With release 2.4.1 the token generation was switched from `UUID` to `SecureRandom`. The secure random byte-array is used as the token value after encoding it with base64.

This can lead to tokens with some special chars, for example `=` or `/`. When a I have a token with `=` and try to send it in an http header, the `=` is treated as a separator and the browser (any maybe also some client libs) removes it. 

### Actual Behavior
The methods `createRefreshToken()` and `createAccessToken()` use the `Base64.encodeBase64` method with default-values for `isChunked` and `urlSafe`:
```
String tokenValue = new String(Base64.encodeBase64(
				DEFAULT_TOKEN_GENERATOR.generateKey()));
```

### Expected Behavior
Use `Base64.encodeBase64( DEFAULT_TOKEN_GENERATOR.generateKey(), true, true )` instead to get a url- and http-header-compatible encoded value.

### Version
spring-security-oauth2 release 2.4.1

See https://github.com/spring-projects/spring-security-oauth/pull/1720/ for the change from `UUID` to `SecureRandom`.

jgrandja:
@torstenkuhnhenne Thank you for reporting this. Would you be interested in submitting a fix for this?

PR #1858, opened 2020-05-05
Use url-safe mode for base64 encoding

torstenkuhnhenne:
Fix for https://github.com/spring-projects/spring-security-oauth/issues/1857.

I am not sure if the class `org.springframework.security.oauth.provider.token.RandomValueProviderTokenServices` also needs this fix.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
@@ -291,7 +291,7 @@ public class DefaultTokenServices implements AuthorizationServerTokenServices, R
 		}
 		int validitySeconds = getRefreshTokenValiditySeconds(authentication.getOAuth2Request());
 		String tokenValue = new String(Base64.encodeBase64(
-				DEFAULT_TOKEN_GENERATOR.generateKey()));
+				DEFAULT_TOKEN_GENERATOR.generateKey(), true, true));
 		if (validitySeconds > 0) {
 			return new DefaultExpiringOAuth2RefreshToken(tokenValue, new Date(System.currentTimeMillis()
 					+ (validitySeconds * 1000L)));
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@torstenkuhnhenne Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1858)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1858) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@torstenkuhnhenne Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1858)!

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/DefaultTokenServices.java
```diff
@@ -291,7 +291,7 @@ private OAuth2RefreshToken createRefreshToken(OAuth2Authentication authenticatio
 		}
 		int validitySeconds = getRefreshTokenValiditySeconds(authentication.getOAuth2Request());
 		String tokenValue = new String(Base64.encodeBase64(
-				DEFAULT_TOKEN_GENERATOR.generateKey()));
+				DEFAULT_TOKEN_GENERATOR.generateKey(), true, true));
```
jgrandja, line 294: Can you also please add a test.

torstenkuhnhenne: Hi @jgrandja, of course I can write a test. Can you give me a tip on what the test should cover?

jgrandja: After looking into this, I realized that it's hard to provide a test since there is setter for overriding `DEFAULT_TOKEN_GENERATOR`. And I'd rather not expose a setter at this point. I think it's safe to not provide a test for this instance.

Also, there were a few failing tests as a result of `isChunked=true` so I updated to call `encodeBase64URLSafe()` instead.

jgrandja:
Thanks for this fix @torstenkuhnhenne. This is now in master and backported to 2.4.x.

torstenkuhnhenne:
You are welcome and thank you @jgrandja for your support and the quick response :-)

PR #1862, opened 2020-05-08
Treat expires_in 0 as not expired

### LINKED ISSUE #1172: Correctly handle when expires_in=0
Some services ( vk.com "vkontakte" for example) responds with "expires_in:0" when requesting access_token witch will never expire (scope "offline" for vk )
In this case module produce OAuth2AccessToken instance with isExpired() instantly "true"

alatushkin:
Fix #1172


alatushkin:
@jgrandja 
here it is

jgrandja:
Thanks for this fix @alatushkin. This is now in master.

Issue #620, opened 2015-11-05
Allow custom AuthenticationProvider on the TokenEndpoint when using @EnableAuthorizationServer


jebeaudet:
When using `@EnableAuthorizationServer`, it imports `AuthorizationServerSecurityConfiguration` which uses `AuthorizationServerSecurityConfigurer` to configure the `/oauth/token`, the `/oauth/token_key` and the `/oauth/check_token` endpoint security properties. A `DaoAuthenticationProvider` coupled with a `ClientDetailsUserDetailsService` is automatically injected with no easy way to configure. 

I'm needing this customization to gracefully apply IP restriction based on OAuth Clients when they authenticate on the `/oauth/token`. The sets of IP will be different depending of the Client. I could inject a `Filter` before the `BasicAuthenticationFilter` but I would like to check the IP restriction once the client has been authenticated. 

The PR #621 propose a solution to this by allowing a custom list of `AuthenticationProvider` to the `AuthorizationServerSecurityConfigurer`.

jebeaudet:
I've added another customization to provide a custom `AuthenticationEventPublisher` on the `AuthenticationManager` instantiated by the `AuthorizationServerSecurityConfigurer`. I've merged the two features in the same PR #621, hopefully it's fine with you and it can get merged upstream soon. If it's a problem, just say so and I'll split them.

jgrandja:
@jebeaudet I'm going to close this based on this [comment](https://github.com/spring-projects/spring-security-oauth/pull/621#pullrequestreview-405460022).

jebeaudet:
Hi @jgrandja 
Unless I'm mistaken, adding an `AuthenticationProvider` in a distinct `WebSecurityConfigurerAdapter` will result in 2 authentication providers in the security chain, my custom one and the `ClientDetailsUserDetailsService` instance wired in `AuthorizationServerSecurityConfigurer`. 

Depending on the order of things, if your custom authentication provider is not used before the default one in the `AuthenticationManager`, it will just not work.

jgrandja:
@jebeaudet 

> Depending on the order of things, if your custom authentication provider is not used before the default one in the `AuthenticationManager`, it will just not work.

Have you tried configuring using `HttpSecurity.authenticationProvider()`? If not, please try it out to determine if it works or not. If it doesn't, then please provide a [minimal sample](https://stackoverflow.com/help/minimal-reproducible-example) demonstrating the issue and I'll take a look at it.

[...]

PR #1865, opened 2020-05-11
Allow custom AuthenticationProvider and AuthenticationEventPublisher

jebeaudet:
Follow up for #620. Thanks


jgrandja:
Thanks for the PR @jebeaudet. This is now in master.

PR #1870, opened 2020-05-25
Specify charset when generating tokens

OrangeDog:
Ensures the token is URL-safe even if the system charset is not a superset of ASCII.


jgrandja:
Thanks for the update @OrangeDog. This is now in master.

OrangeDog:
What's the reason for switching to `Charset.forName("US-ASCII")`?

jgrandja:
There was a build failure:

> [INFO] --- animal-sniffer-maven-plugin:1.6:check (enforce-java-6) @ spring-security-oauth2 ---
[INFO] Checking unresolved references to org.codehaus.mojo.signature:java16:1.0
[ERROR] Undefined reference: java/nio/charset/StandardCharsets.US_ASCII#Ljava/nio/charset/Charset; in /Users/jgrandja/workspace/spring-security-oauth-fork/spring-security-oauth2/target/classes/org/springframework/security/oauth2/provider/token/DefaultTokenServices.class

OrangeDog:
Ok, I didn't realise this was supposed to still support Java 6.

PR #1907, opened 2021-02-11
Fix NPE in JdbcTokenStore

MikeN123:
If an old Authentication failed to deserialize (e.g., after upgraded Spring Security core libs) during the creation of a token, a NPE would be thrown as `readAuthentication` return `null` in that case. This should fix that.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/JdbcTokenStore.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/JdbcTokenStore.java
@@ -129,12 +129,14 @@ public class JdbcTokenStore implements TokenStore {
 			LOG.error("Could not extract access token for authentication " + authentication, e);
 		}
 
-		if (accessToken != null
-				&& !key.equals(authenticationKeyGenerator.extractKey(readAuthentication(accessToken.getValue())))) {
-			removeAccessToken(accessToken.getValue());
-			// Keep the store consistent (maybe the same user is represented by this authentication but the details have
-			// changed)
-			storeAccessToken(accessToken, authentication);
+		if (accessToken != null) {
+			OAuth2Authentication oldAuthentication = readAuthentication(accessToken.getValue());
+			if (oldAuthentication == null || !key.equals(authenticationKeyGenerator.extractKey(oldAuthentication))) {
+				removeAccessToken(accessToken.getValue());
+				// Keep the store consistent (maybe the same user is represented by this authentication but the details have
+				// changed)
+				storeAccessToken(accessToken, authentication);
+			}
 		}
 		return accessToken;
 	}
--- a/spring-security-oauth2/src/test/java/org/springframework/security/oauth2/provider/token/store/JdbcTokenStoreTests.java
+++ b/spring-security-oauth2/src/test/java/org/springframework/security/oauth2/provider/token/store/JdbcTokenStoreTests.java
@@ -15,6 +15,7 @@ import org.springframework.jdbc.datasource.embedded.EmbeddedDatabase;
 import org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseBuilder;
 import org.springframework.security.oauth2.common.DefaultOAuth2AccessToken;
 import org.springframework.security.oauth2.common.OAuth2AccessToken;
+import org.springframework.security.oauth2.common.util.DefaultSerializationStrategy;
 import org.springframework.security.oauth2.common.util.SerializationStrategy;
 import org.springframework.security.oauth2.common.util.SerializationUtils;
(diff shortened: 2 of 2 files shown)


jgrandja:
Thanks for the PR @MikeN123. Can you please provide a test that demonstrates the fix resolves the issue.

MikeN123:
That's quite difficult, as in our case it was triggered by an outdated serialized version of a class, which caused:

```
java.lang.IllegalArgumentException: java.io.InvalidClassException: org.springframework.security.core.authority.SimpleGrantedAuthority; local class incompatible: stream classdesc serialVersionUID = 530, local class serialVersionUID = 540
```

Of course I can fake some kind of a deserialization problem, but it won't really be the real thing...
Suggestions on how to approach this?

MikeN123:
I have added a test that writes an invalid serialized version of the authentication object to the database and then tries to read it back. This test fails with the old code and succeeds with the new code. It does log a WARN though, but that is unavoidable as that is always logged if the deserialization fails.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/JdbcTokenStore.java
```diff
@@ -129,12 +129,14 @@ public OAuth2AccessToken mapRow(ResultSet rs, int rowNum) throws SQLException {
...
 
-		if (accessToken != null
-				&& !key.equals(authenticationKeyGenerator.extractKey(readAuthentication(accessToken.getValue())))) {
-			removeAccessToken(accessToken.getValue());
-			// Keep the store consistent (maybe the same user is represented by this authentication but the details have
-			// changed)
-			storeAccessToken(accessToken, authentication);
+		if (accessToken != null) {
+			OAuth2Authentication oldAuthentication = readAuthentication(accessToken.getValue());
+			if (oldAuthentication == null || !key.equals(authenticationKeyGenerator.extractKey(oldAuthentication))) {
```
jgrandja, line 134: @MikeN123 I'm feeling like we should not apply this change. It hides the fact that the persistent store contains invalid / incompatible data. Given that the Spring Security libraries have been upgraded, it is expected that these deserialization errors would occur if the stored version is incompatible and therefore should be revealed to the application via an error instead of hiding it altogether.

I believe some form of cleanup or migration would need to happen to the persistent store if Spring Security libraries are upgraded.

jgrandja (changes requested):
Thanks for the PR @MikeN123. Please see review comment.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/JdbcTokenStore.java
MikeN123: I disagree. There are all kinds of places where the project already does automatic cleanup. It even does that in this place - `readAuthentication()` will silently remove the stored data, but log a warning, and the next request will succeed. The fact that a NPE is thrown is just an error, in other places it will happily continue (e.g., earlier in this method it retrieves `accessToken`, if `deserializeAccessToken()` fails there, it returns `null` as well, which is subsequently checked and not throwing some kind of error.

It is undoable as a developer to clean up the persistent store at the right times, as it is almost impossible to keep track of when Spring Security does or does not make a change that changes the serialization version of their classes (especially when you pull in a Spring Security upgrade through a Spring Boot upgrade).

If it should be doing something different here, then at least it should just return `null` on deserialization failure, like it does when the result of `selectAccessTokenFromAuthenticationSql` fails to deserialize. Otherwise it's just inconsistent with all other places where deserialization failures silently remove the stored data.

There is also not really something that the application can do about such failures, except for truncating the store. This is used in the `token` endpoint which is directly exposed to the outside world, it is not like the library user could be doing some error handling themselves.

MikeN123:
I've replied to your comment (commenting here so the bot can do his job and update the labels).

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/JdbcTokenStore.java
jgrandja: @MikeN123 Apologies as my initial assessment of the changes was incorrect. This looks good and I'll have it merged.

jgrandja (approved):

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/store/JdbcTokenStore.java
MikeN123: Np, thanks :)

jgrandja:
Thanks for this fix @MikeN123 ! This is now in master and 2.4.x.

PR #1910, opened 2021-02-18
Add clock skew support to AccessTokenProviderChain

### LINKED ISSUE #1909: Missing clock skew support in AccessTokenProviderChain
### Summary

The fix for #1287 added a clock skew to `OAuth2RestTemplate` but there is another expiration check in `AccessTokenProviderChain` that does not take this clock skew into account which renders this fix useless when used in combination with `AccessTokenProviderChain`.

### Version

2.5.0.RELEASE

akloeber:
This fixes #1909 and adds clock skew support to AccessTokenProviderChain. As with OAuth2RestTemplate the default clock skew is set to 30s.

Background: The fix for #1287 added a clock skew to OAuth2RestTemplate but there is another expiration check in AccessTokenProviderChain that does not take this into account which renders this fix useless when used in combination with AccessTokenProviderChain.

### DIFF
--- a/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/OAuth2RestTemplate.java
+++ b/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/OAuth2RestTemplate.java
@@ -2,6 +2,7 @@ package org.springframework.security.oauth2.client;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.lang.reflect.Field;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URLEncoder;
@@ -26,6 +27,7 @@ import org.springframework.security.oauth2.common.AuthenticationScheme;
 import org.springframework.security.oauth2.common.OAuth2AccessToken;
 import org.springframework.security.oauth2.common.exceptions.InvalidTokenException;
 import org.springframework.util.Assert;
+import org.springframework.util.ReflectionUtils;
 import org.springframework.web.client.RequestCallback;
 import org.springframework.web.client.ResponseErrorHandler;
 import org.springframework.web.client.ResponseExtractor;
@@ -287,6 +289,7 @@ public class OAuth2RestTemplate extends RestTemplate implements OAuth2RestOperat
 
 	public void setAccessTokenProvider(AccessTokenProvider accessTokenProvider) {
 		this.accessTokenProvider = accessTokenProvider;
+		propagateClockSkewToAccessTokenProvider(clockSkew, accessTokenProvider);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@akloeber Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1910)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1910) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@akloeber Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security-oauth&pullRequestId=1910)!

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/token/OAuth2AccessTokenExpiryChecker.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Calendar;
+
+/**
+ * Checks expiration of an OAuth2AccessToken while also taking a clock skew into account.
+ *
+ * @author Andreas Klöber
+ * @deprecated See the <a href="https://github.com/spring-projects/spring-security/wiki/OAuth-2.0-Migration-Guide">OAuth 2.0 Migration Guide</a> for Spring Security 5.
+ */
+@Deprecated
+public class OAuth2AccessTokenExpiryChecker {
```
jgrandja, line 14: We cannot add public API's going forward as of 2.5.0 as it was the final minor release given this project is coming to EOL - see [announcement](https://spring.io/blog/2020/05/07/end-of-life-for-spring-security-oauth). The changes in this PR need to be patch changes.

Here is a suggestion for the patch change:

- Introduce `private int clockSkew = 30` in `AccessTokenProviderChain` - same as `OAuth2RestTemplate`
- When `OAuth2RestTemplate.setClockSkew()` is called, it will also set `AccessTokenProviderChain.clockSkew` member to the same value using reflection. Take a look at this [test](https://github.com/spring-projects/spring-security-oauth/blob/304c43bed142853442d1a76e5969e337cc887378/spring-security-oauth2/src/test/java/org/springframework/security/oauth2/provider/token/store/jwk/JwkTokenStoreTests.java#L63) to see how you can leverage `ReflectionUtils`.

I know this is not a pretty fix but it will work for the patch.

jgrandja (changes requested):
Thanks for the PR @akloeber. Please see review comment.

spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/token/OAuth2AccessTokenExpiryChecker.java
akloeber: @jgrandja Thanks for the feedback. I only introduced `OAuth2AccessTokenExpiryChecker` class as I did not find a better place to share the expiration-check-logic.  I'll have a look at your suggestion.

jgrandja:
Thanks for this fix @akloeber ! This is now in master.

akloeber:
@jgrandja Sorry to bother you again, but what's the roadmap for the next 2.5.1 release? We want to consume this bugfix in our application landscape to stop these issues.

jgrandja:
@akloeber I scheduled `2.5.1` for April 8.

jgrandja:
@akloeber `2.5.1` is now available in Central

PR #1941, opened 2021-10-13
Polish logs

jzheaux:


jgrandja:
Thanks for the PR @jzheaux. This is now in main.

